PR #6, opened 2019-07-11
feat(tls) new method resty.kong.tls.upstream_cert_and_key for

dndx:
dynamically overriding client certificate used while connecting to
upstream.

Rendered doc: https://github.com/Kong/lua-kong-nginx-module/blob/feat/upstream_cert/README.md#restykongtlsupstream_cert_and_key

Ran tests under Valgrind and leak test mode, no issues.


README.md
```diff
@@ -83,6 +100,31 @@ In this case caller should use
 associate this session with one of the previous handshakes to identify the connecting
 client.
 
+[Back to TOC](#table-of-contents)
+
+resty.kong.tls.set\_upstream\_cert\_and\_key
+---------------------------------------
```
thibaultcha: style: I notice we fixed the style of the above `resty.kong.tls.get_full_client_certificate_chain` header, however this header isn't properly styled - not a blocker for me, just pointing it out since I see the above fix.

lualib/resty/kong/tls.lua
```diff
@@ -131,4 +154,38 @@ do
...
+        ['access'] = true,
+    }
+
+    function _M.set_upstream_cert_and_key(chain, key)
+        if not ALLOWED_PHASES[get_phase()] then
+            error("API disabled in the current context", 2)
+        end
+
+        if not chain or not key then
+            error("chain and key must not be nil")
```
thibaultcha: This call is missing the stack level argument

patches/1.15.8.1/nginx-1.15.8-upstream_client_certificate.patch
```diff
@@ -0,0 +1,25 @@
...
+         }
+     }
+ 
++#if (NGX_HTTP_LUA_KONG)
++    ngx_http_lua_kong_set_upstream_ssl(r, c);
++#endif
++
+     r->connection->log->action = "SSL handshaking to upstream";
+ 
+     rc = ngx_ssl_handshake(c);
```
thibaultcha: I believe that this patch should belong to our openresty-patches repository? It's not going to be a good thing if we split them between openresty-patches and this repository (also consider we have a few non Kong-related ones in openresty-build-tools as well).

Also, this patch should, if possible, set `HAVE_KONG_SET_UPSTREAM_SSL` (or similar), which, if not set, would cause `ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key()` to return `NGX_DECLINED` (or similar), which, in turn, would cause `kong.tls.set_upstream_cert_and_key()` to return an error basically saying the feature is "unsupported because of missing patch xyz".

src/ngx_http_lua_kong_module.c
```diff
@@ -305,3 +305,172 @@ ngx_http_lua_kong_ffi_get_full_client_certificate_chain(ngx_http_request_t *r,
...
+            return NGX_ERROR;
+        }
+
+        cln->data = ctx;
+        cln->handler = ngx_http_lua_kong_cleanup;
+
+        ngx_http_set_ctx(r, ctx, ngx_http_lua_kong_module);
+
+    } else {
+        if (ctx->upstream_client_certificate_chain != NULL) {
```
thibaultcha: Could we factorize this into `} elseif { ... }`?

src/ngx_http_lua_kong_module.c
```diff
@@ -305,3 +305,172 @@ ngx_http_lua_kong_ffi_get_full_client_certificate_chain(ngx_http_request_t *r,
...
+    return X509_up_ref(in) == 0 ? NULL : in;
+}
+
+
+static void
+ngx_http_lua_kong_cleanup(void *data)
+{
+    ngx_http_lua_kong_ctx_t     *ctx = data;
+
+    if (ctx->upstream_client_certificate_chain) {
```
thibaultcha: An explicit check for `!= NULL` maybe? We do so below in `ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key()`

src/ngx_http_lua_kong_module.c
```diff
@@ -305,3 +305,172 @@ ngx_http_lua_kong_ffi_get_full_client_certificate_chain(ngx_http_request_t *r,
...
+            ngx_http_lua_kong_cleanup(ctx);
+
+            ctx->upstream_client_certificate_chain = NULL;
+            ctx->upstream_client_private_key = NULL;
+        }
+    }
+
+    if (chain == NULL || key == NULL) {
+        return NGX_ERROR;
+    }
```
thibaultcha: It sounds to me like we could move this branch all the way up in this function's body? Or is the intent to clear our `ngx_http_lua_kong_ctx_t` even when the arguments to this function are invalid (which may sound like an undesired side-effect)? In any case, this shouldn't happen due to our safeguards in the Lua-land, but it does make the reader's life a bit easier if it could be moved up, I would say.
I am fine either way, just bringing it to your attention just in case.

src/ngx_http_lua_kong_module.c
```diff
@@ -305,3 +305,172 @@ ngx_http_lua_kong_ffi_get_full_client_certificate_chain(ngx_http_request_t *r,
     return NGX_ABORT;
 #endif
 }
+
+
+#if (NGX_HTTP_SSL)
```
thibaultcha: It was my understanding that you wanted to get rid of those? Nonetheless, if we do decide to keep them around, the other FFI C functions in this module have safeguards for when NGINX is compiled without SSL support. Or were those safeguards what you wanted to get rid of (and not the macro guards)?

README.md
```diff
@@ -57,8 +72,10 @@ This function returns `true` when the call is successful. Otherwise it returns
 `nil` and a string describing the error.
 
 
+[Back to TOC](#table-of-contents)
```
thibaultcha: A very minor comment: there's two line jumps before this link. Really not a blocker.

lualib/resty/kong/tls.lua
```diff
@@ -131,4 +154,38 @@ do
...
+        end
+
+        local r = get_request()
+        size_ptr[0] = DEFAULT_CERT_CHAIN_SIZE
```
thibaultcha: It doesn't look like we need this allocation?

lualib/resty/kong/tls.lua
kikito: This will be triggered when either `chain` or `key` are `false` too. That could be fixed by changing the error message to `chain and key must be present`.

However what happens if `chain` or `key` is something more ... exotic, like a table value? Someone could do `:set_upstream_cert_and_key` instead of `.set_upstream_cert_and_key` for example.

Is `C.ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key` able to handle such errors? If not, I suggest changing this to a more explicit check for types.

What about empty strings? Are they considered valid keys?

```
if key == "" or type(key) ~= "string" then
  error("key must be a non-empty string")
end

if type(chain) ~= ... then -- not sure about this one, don't know what STACK_OF(X509) means
  ...
end
```

thibaultcha: In `resty.*` module,s arguments validation is typically kept to a minimum. We could have stricter arguments validation in our `kong.*` PDK module though.

src/ngx_http_lua_kong_module.c
dndx: Good catch. Changed.

src/ngx_http_lua_kong_module.c
dndx: I do intend to remove them. But for the scope of this feature I want to keep the PR more readable so they were left in. I would have a separate PR that removes all of them (but the intention is to get rid of them soon).

lualib/resty/kong/tls.lua
dndx: And PDK does https://github.com/Kong/kong/pull/4797/files#diff-0a162c8833b46ddf83218f6afa5a32cdR452! Although we are limited about what kind of validation we can do. If the user passed the wrong type of `cdata` pointer in we have no way to tell. But that's the same situation with `ngx.ssl` right now as well.

lualib/resty/kong/tls.lua
dndx: Copy-n-paste error. Thanks for pointing out!

src/ngx_http_lua_kong_module.c
thibaultcha: @dndx Ok! Thanks for clarifying

patches/1.15.8.1/nginx-1.15.8-upstream_client_certificate.patch
dndx: Okay, I will move the patch to the `openresty-patches` repo.

For the macro guard, I discussed with @thibaultcha and here is what we will do:

https://trello.com/c/V88fMBlQ/1883-add-patch-and-nginx-module-checks-for-kong-core

It falls outside of this PR's scope though. So let's merge this PR first and open another PR to do that.

dndx:
Patches has been moved over to https://github.com/Kong/openresty-patches/pull/46. Merging this.

PR #8, opened 2019-09-07
fix(Makefile) solve the problem of tls.lua placement path error

Bevisy:
The default Makefile can't install tls.lua to correct package.path.

Refer to the latest container image of kong1.3.0 version, modify Makfile to solve the problem of tls.lua placement path error.

Signed-off-by: bevisy <binbin36520@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 OPENRESTY_PREFIX=/usr/local/openresty
 
 #LUA_VERSION := 5.1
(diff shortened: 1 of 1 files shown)


chronolaw:
We can use env `LUA_LIB_DIR` to get the same result.

Just like `make install LUA_LIB_DIR=path/to/openresty/lualib`.

PR #11, opened 2020-01-15
feat(tls) new methods to override upstream ssl verify behaviour

fffonion:
See also https://github.com/Kong/kong-build-tools/pull/206

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ Table of Contents
     * [resty.kong.tls.disable\_session\_reuse](#restykongtlsdisable_session_reuse)
     * [resty.kong.tls.get\_full\_client\_certificate\_chain](#restykongtlsget_full_client_certificate_chain)
     * [resty.kong.tls.set\_upstream\_cert\_and\_key](#restykongtlsset_upstream_cert_and_key)
+    * [resty.kong.tls.set\_upstream\_ssl\_trusted\_store](#restykongtlsset_upstream_ssl_trusted_store)
+    * [resty.kong.tls.set\_upstream\_ssl\_verify](#restykongtlsset_upstream_ssl_verify)
+    * [resty.kong.tls.set\_upstream\_ssl\_verify\_depth](#restykongtlsset_upstream_ssl_verify_depth)
     * [resty.kong.tls.disable\_proxy\_ssl](#restykongtlsdisable_proxy_ssl)
 * [License](#license)
 
@@ -134,6 +137,106 @@ previous ones.
 
 [Back to TOC](#table-of-contents)
 
+resty.kong.tls.set\_upstream\_ssl\_trusted\_store
+--------------------------------------------
+**syntax:** *ok, err = resty.kong.tls.set\_upstream\_ssl\_trusted\_store(store)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, balancer_by_lua&#42;*
+
+**subsystems:** *http*
+
+Set upstream ssl verification trusted store of current request. Global setting set by
+`proxy_ssl_trusted_certificate` will be overwritten for the current request.
+
+`store` is a `X509_STORE*` cdata that can be created by
+[resty.openssl.x509.store.new](https://github.com/fffonion/lua-resty-openssl#storenew).
+
+On success, this function returns `true` and future handshakes with upstream servers
+will be verified with given store. Otherwise `nil` and a string describing the
+error will be returned.
+
+This function can be called multiple times in the same request. Later calls override
+previous ones.
+
+Example:
+```lua
+local x509 = require("resty.openssl.x509")
+local crt, err = x509.new([[-----BEGIN CERTIFICATE-----
+...
+-----END CERTIFICATE-----]])
+if err then
+    ngx.log(ngx.ERR, "failed to parse cert: ", err)
+    ngx.exit(500)
+end
+local store = require("resty.openssl.x509.store")
+local st, err = store.new()
+if err then
+    ngx.log(ngx.ERR, "failed to create store: ", err)
+    ngx.exit(500)
+end
+local ok, err = st:add(crt)
+if err then
+    ngx.log(ngx.ERR, "failed to add cert to store: ", err)
+    ngx.exit(500)
+end
+-- st:add can be called multiple times, also accept a crl
+-- st:add(another_crt)
+-- st:add(crl)
+-- OR
+-- st:use_default() to load default CA bundle
+local tls = require("resty.kong.tls")
+local ok, err = tls.set_upstream_ssl_trusted_store(st.ctx)
+if err then
+    ngx.log(ngx.ERR, "failed to set upstream trusted store: ", err)
+    ngx.exit(500)
+end
+local ok, err = tls.set_upstream_ssl_verify(true)
+if err then
+    ngx.log(ngx.ERR, "failed to set upstream ssl verify: ", err)
+    ngx.exit(500)
+end
+```
+
+[Back to TOC](#table-of-contents)
+
+resty.kong.tls.set\_upstream\_ssl\_verify
+--------------------------------------------
+**syntax:** *ok, err = resty.kong.tls.set\_upstream\_ssl\_verify(verify)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, balancer_by_lua&#42;*
+
+**subsystems:** *http*
+
+Set upstream ssl verification enablement of current request to the given boolean
+argument `verify`. Global setting set by `proxy_ssl_verify` will be overwritten.
+
+On success, this function returns `true`. Otherwise `nil` and a string
+describing the error will be returned.
+
+This function can be called multiple times in the same request. Later calls override
+previous ones.
+
+[Back to TOC](#table-of-contents)
+
+resty.kong.tls.set\_upstream\_ssl\_verify\_depth
+--------------------------------------------
+**syntax:** *ok, err = resty.kong.tls.set\_upstream\_ssl\_verify\_depth(depth)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, balancer_by_lua&#42;*
+
+**subsystems:** *http*
+
+Set upstream ssl verification depth of current request to the given non-negative integer
+argument `depth`. Global setting set by `proxy_ssl_verify_depth` will be overwritten.
+
+On success, this function returns `true`. Otherwise `nil` and a string
+describing the error will be returned.
+
+This function can be called multiple times in the same request. Later calls override
+previous ones.
+
+[Back to TOC](#table-of-contents)
+
 resty.kong.tls.disable\_proxy\_ssl
 ----------------------------------
 **syntax:** *ok, err = resty.kong.tls.disable_proxy_ssl()*
@@ -172,4 +275,3 @@ limitations under the License.
 ```
 
 [Back to TOC](#table-of-contents)
-
--- a/src/ngx_http_lua_kong_module.c
+++ b/src/ngx_http_lua_kong_module.c
@@ -325,6 +325,7 @@ ngx_http_lua_kong_set_upstream_ssl(ngx_http_request_t *r, ngx_connection_t *c)
     STACK_OF(X509)              *chain;
     EVP_PKEY                    *pkey;
     X509                        *x509;
+    X509_STORE                  *store;
 #ifdef OPENSSL_IS_BORINGSSL
     size_t                       i;
 #else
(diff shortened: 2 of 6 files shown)


fffonion:
The X509_STORE construstor needs lua-resty-openssl >=0.5.0 (unreleased). Ideally there can be a helper function like how `ssl.parse_pem_cert` is done to create the X509_STORE struct needed by `set_upstream_ssl_trusted_store`.

src/ngx_http_lua_kong_module.c
```diff
@@ -325,23 +325,31 @@ ngx_http_lua_kong_set_upstream_ssl(ngx_http_request_t *r, ngx_connection_t *c)
...
 #ifdef OPENSSL_IS_BORINGSSL
     size_t                       i;
 #else
     int                          i;
 #endif
 
     ctx = ngx_http_get_module_ctx(r, ngx_http_lua_kong_module);
-    if (ctx == NULL || ctx->upstream_client_certificate_chain == NULL) {
+
+    if (ctx == NULL || (ctx->upstream_client_certificate_chain == NULL &&
```
dndx: Since now we can set either client cert or verify param, this check can be changed to check for `ctx == NULL` only, and use `if` branch below to distinguish between those.

src/ngx_http_lua_kong_module.c
```diff
@@ -325,23 +325,31 @@ ngx_http_lua_kong_set_upstream_ssl(ngx_http_request_t *r, ngx_connection_t *c)
...
+                    "skip overriding upstream SSL configuration, "
+                    "certificate and trusted store not set");
         return;
     }
 
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, c->log, 0,
                    "overriding upstream SSL configuration");
 
+    if (ctx->upstream_client_certificate_chain == NULL) {
+        goto skip_client_cert_key;
```
dndx: We should avoid using `goto` if possible. A simple `if ctx->upstream_client_certificate_chain != NULL` should work well for that case.

src/ngx_http_lua_kong_module.c
```diff
@@ -325,23 +325,31 @@ ngx_http_lua_kong_set_upstream_ssl(ngx_http_request_t *r, ngx_connection_t *c)
...
+                        ctx->upstream_ssl_verify_depth == 0)) {
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
-                       "skip overriding upstream SSL configuration, "
-                       "certificate not set");
+                    "skip overriding upstream SSL configuration, "
+                    "certificate and trusted store not set");
         return;
     }
 
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, c->log, 0,
```
dndx: This should be moved into each `if` branch as necessary, as the structure of the function changes a lot because of the addition.

src/ngx_http_lua_kong_module.c
```diff
@@ -383,6 +391,22 @@ ngx_http_lua_kong_set_upstream_ssl(ngx_http_request_t *r, ngx_connection_t *c)
...
+    if (ctx->upstream_trusted_store != NULL) {
+        store = ctx->upstream_trusted_store;
+
+        if (SSL_set1_verify_cert_store(sc, store) == 0) {
+            ngx_ssl_error(NGX_LOG_ALERT, c->log, 0, "SSL_set1_verify_cert_store() failed");
+            goto failed;
+        }
+    }
+
+    if (ctx->upstream_ssl_verify_depth > 0) {
```
dndx: This doesn't seems right. Nginx and OpenSSL both accepts `verify depth = 0` as an acceptable value, therefore we should not use `0` as the "not set flag".

The typical two ways in Nginx of handling this:

1. Add a bitfield `ngx_uint_t upstream_ssl_verify_depth;` at the end. Make a comment to indicate that it can be a 1-bit bitfield. Since no other bit-field is present, making it a single bit right now causes unnecessary `AND` operation when reading so it is indicated in the comment only. When the next bit flag got added it will be changed to a bit field. (See http://lxr.nginx.org/source/src//core/ngx_cycle.h#0119, `ngx_uint_t                transparent;  /* unsigned  transparent:1; */` definition as an example)
2. Simply use `ngx_int_t` for the depth and use `-1` as the "unset" value

Personally 1 is my favorite. It's a bit more work, but safer, as the zeroed `ctx` allocated by `pcalloc` is already in the  "unset" state. For 2 `ctx->upstream_ssl_verify_depth`must be explicitly initialized to `-1` after allocation.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
         sk_X509_pop_free(ctx->upstream_client_certificate_chain, X509_free);
         EVP_PKEY_free(ctx->upstream_client_private_key);
     }
+
+    if (ctx->upstream_trusted_store != NULL) {
```
dndx: We should rewrite this function to use the `ngx_http_lua_kong_cleanup_cert_and_key` and `ngx_http_lua_kong_cleanup_trusted_store` helper instead.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
...
+        X509_STORE_free(ctx->upstream_trusted_store);
+    }
 }
 
+static void
+ngx_http_lua_kong_cleanup_cert_and_key(void *data)
```
dndx: This function should be written to accept `ngx_http_lua_kong_ctx_t` instead of `void`, as it is not invoked by the cleanup handler directly and no need to be defined as `void`.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
...
+        sk_X509_pop_free(ctx->upstream_client_certificate_chain, X509_free);
+        EVP_PKEY_free(ctx->upstream_client_private_key);
+    }
+}
 
-int
-ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
-    void *_chain, void *_key)
+static void
+ngx_http_lua_kong_cleanup_trusted_store(void *data)
```
dndx: Same here, use `ngx_http_lua_kong_ctx_t` instead.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
...
+    ngx_http_lua_kong_ctx_t     *ctx = data;
 
-    if (chain == NULL || key == NULL) {
-        return NGX_ERROR;
+    if (ctx->upstream_trusted_store != NULL) {
+        X509_STORE_free(ctx->upstream_trusted_store);
     }
+}
+
+ngx_http_lua_kong_ctx_t *
```
dndx: Make this `static`, as it is only used within this file.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
...
-            return NGX_ERROR;
+            return NULL;
         }
 
         cln->data = ctx;
         cln->handler = ngx_http_lua_kong_cleanup;
 
         ngx_http_set_ctx(r, ctx, ngx_http_lua_kong_module);
+    }
+    return ctx;
```
dndx: Style: add a newline before this.

src/ngx_http_lua_kong_module.c
```diff
@@ -407,42 +431,78 @@ ngx_http_lua_kong_cleanup(void *data)
...
+    STACK_OF(X509)              *new_chain;
+    ngx_http_lua_kong_ctx_t     *ctx;
+
+    if (chain == NULL || key == NULL) {
+        return NGX_ERROR;
+    }
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx == NULL) {
+        return NGX_ERROR;
```
dndx: Style: add a newline after this line. Nginx C source code always add a newline before `else if` or `else`.

src/ngx_http_lua_kong_module.c
```diff
@@ -471,4 +531,91 @@ ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
...
+    X509_STORE                  *store = _store;
+    ngx_http_lua_kong_ctx_t     *ctx;
+
+    if (store == NULL) {
+        return NGX_ERROR;
+    }
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx == NULL) {
+        return NGX_ERROR;
```
dndx: Same here for the style.

src/ngx_http_lua_kong_module.c
```diff
@@ -471,4 +531,91 @@ ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
...
+    int verify, int depth)
+{
+    ngx_http_lua_kong_ctx_t     *ctx;
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx == NULL) {
+        return NGX_ERROR;
+    }
+
+    ctx->upstream_ssl_verify = verify ?
```
dndx: Instead of doing bit shift, we can use `NGX_CONF_UNSET` http://lxr.nginx.org/source/src/core/ngx_conf_file.h#0056 as the marker for unset value. This is very common usage within Nginx itself.

See: http://lxr.nginx.org/source/src/http/modules/ngx_http_autoindex_module.c#1031 as an example (`enable` flag).

src/ngx_http_lua_kong_module.c
```diff
@@ -471,4 +531,91 @@ ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
...
+    ngx_http_lua_kong_ctx_t     *ctx;
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx == NULL) {
+        return NGX_ERROR;
+    }
+
+    /* ssl_verify_depth can be set to 0 as well (self-signed only), add a marker
+    * so later we know this flag is modified */
+    ctx->upstream_ssl_verify_depth = depth << 1 | 0x1;
```
dndx: Same here, we can use `NGX_CONF_UNSET_UINT` http://lxr.nginx.org/source/src/core/ngx_conf_file.h#0057 instead.

src/ngx_http_lua_kong_module.c
```diff
@@ -471,4 +531,91 @@ ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
...
+
+    return NGX_OK;
+}
+
+ngx_uint_t
+ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
+    int proxy_ssl_verify)
```
dndx: This should be `ngx_flag_t` instead to match http://lxr.nginx.org/source/src/http/ngx_http_upstream.h#0236. We only use unqualified `int` for FFI C APIs. For Nginx C function they are always defined using the appropriate Nginx types.

t/002-upstream-tls.t
```diff
@@ -277,3 +277,441 @@ verify:1, error:0, depth:0, subject:"/C=US/ST=California/O=Kong Testing/CN=foo@e
...
+
+            -- verify depth 0 means only self signed cert is allowed
+            local ok, err = tls.set_upstream_ssl_verify_depth(0)
+            if not ok then
+                ngx.say("set_upstream_ssl_verify_depth failed: ", err)
+            end
+        }
+        proxy_ssl_trusted_certificate ../../cert/ca.crt;
+        proxy_ssl_verify on;
+        proxy_ssl_verify_depth 1; # default value
```
dndx: Set this to `100` just to be sure that the Lua API indeed works....

t/002-upstream-tls.t
```diff
@@ -277,3 +277,441 @@ verify:1, error:0, depth:0, subject:"/C=US/ST=California/O=Kong Testing/CN=foo@e
...
+            local tls = require("resty.kong.tls")
+
+            local ok, err = tls.set_upstream_ssl_verify_depth(2)
+            if not ok then
+                ngx.say("set_upstream_ssl_verify_depth failed: ", err)
+            end
+        }
+        proxy_ssl_trusted_certificate ../../cert/ca.crt;
+        proxy_ssl_verify on;
+        proxy_ssl_verify_depth 1; # default value
```
dndx: Same, set this to `0` to ensure it works.

t/002-upstream-tls.t
```diff
@@ -277,3 +277,441 @@ verify:1, error:0, depth:0, subject:"/C=US/ST=California/O=Kong Testing/CN=foo@e
...
+
+--- error_code: 200
+--- no_error_log
+[error]
+[crit]
+[alert]
+
+
+
+=== TEST 11: setting insufficient trusted store with tls.set_upstream_ssl_trusted_store, verify failed
```
dndx: This should be called "incorrect trusted store" instead? "insufficient" sounds a little strange...

README.md
```diff
@@ -134,6 +137,82 @@ previous ones.
...
+previous ones.
+
+Example:
+```lua
+local tls = require("resty.kong.tls")
+local x509 = require("resty.openssl.x509")
+local store = require("resty.openssl.x509.store")
+local st = assert(store.new())
+assert(st:add([[-----BEGIN CERTIFICATE-----
+...
```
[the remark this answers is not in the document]
fffonion: Actually this syntax is wrong, store:add will expect a x509 instance. I'll fix in readme.

src/ngx_http_lua_kong_module.c
fffonion: Discussed IRL and took approach 1 as path forward.

src/ngx_http_lua_kong_module.h
```diff
@@ -10,11 +10,20 @@
 typedef struct {
     STACK_OF(X509)      *upstream_client_certificate_chain;
     EVP_PKEY            *upstream_client_private_key;
+    X509_STORE          *upstream_trusted_store;
+    ngx_uint_t          upstream_ssl_verify;
```
dndx: This can be moved to a bit field below as well.

src/ngx_http_lua_kong_module.h
```diff
@@ -10,11 +10,20 @@
...
+    ngx_uint_t          upstream_ssl_verify_depth;
+    unsigned            upstream_ssl_verify_set:1,
+                        upstream_ssl_verify_depth_set:1;
```
dndx: Style: add `unsigned` before the declaration:

`unsigned upstream_ssl_verify_depth_set:1;`

src/ngx_http_lua_kong_module.c
```diff
@@ -471,4 +530,90 @@ ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
...
+        return NGX_ERROR;
+    }
+
+    ctx->upstream_ssl_verify_depth_set = 1;
+    ctx->upstream_ssl_verify_depth = depth;
+
+    return NGX_OK;
+}
+
+ngx_uint_t
```
dndx: This should return `ngx_flag_t` instead.

PR #13, opened 2020-05-13
feat(balancer) recreate request to update Host header

locao:


hutchic:
will these changes be compatible with past versions of Kong?

src/ngx_http_lua_kong_module.c
```diff
@@ -624,3 +713,562 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
 }
 
 #endif
+
+
+extern ngx_module_t ngx_http_proxy_module;
+
+
+static ngx_int_t
+ngx_http_proxy_body_output_filter(void *data, ngx_chain_t *in)
```
fffonion, line 722: for readability i'd suggest move this function and its companion structs that we copy paste from nginx to a separate file, and also add a macro check on the nginx core version we based off. since this function may change in the nginx core and the created request may out of sync

lualib/resty/kong/balancer.lua
```diff
@@ -0,0 +1,59 @@
...
+        end
+
+        local r = get_request()
+
+        local ret = C.ngx_http_lua_kong_ffi_recreate_request(r)
+        if ret == ngx.NGX_OK then
+            return true
+        end
+
+        return nil, "could not recreate request"
```
fffonion, line 52: in this case, should we just `error()` out and fail the request?

src/ngx_http_lua_kong_module.c
```diff
@@ -624,3 +713,562 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+            // keep the end of the Host header name position
+            orig_pos = e.pos;
+            // run the script
+            while (*(uintptr_t *) e.ip) {
+                code = *(ngx_http_script_code_pt *) e.ip;
+                code((ngx_http_script_engine_t *) &e);
+            }
+            // overwrite the Host header value
+            e.pos = orig_pos;
+            e.pos = ngx_copy(e.pos, r->headers_in.host->value.data, r->headers_in.host->value.len);
```
fffonion, line 1160: if i understand correctly, `e.pos` is point to buffer `b` which is created from Line 1054 with the length calculated from known headers. if we overwrite the header larger than the previous, this is going to overflow. (let's also make the test a HTTP 1.0 request in test to force nginx to parse all CRLF correctly and validate all sent headers are intact)

src/ngx_http_lua_kong_module.c
fffonion: this will also make it possible to support different versions of kong if we want to do that

lualib/resty/kong/balancer.lua
```diff
@@ -0,0 +1,59 @@
...
+
+    else
+        local getfenv = getfenv
+
+        function get_request()
+            return getfenv(0).__ngx_req
+        end
+    end
+end
+
```
bungle, line 37: Why not just:

```
local get_request = require "resty.core.base".get_request
```

src/ngx_http_lua_kong_module.c
```diff
@@ -624,3 +713,562 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+        }
+    }
+
+
+    b = ngx_create_temp_buf(r->pool, len);
+    if (b == NULL) {
+        return NGX_ERROR;
+    }
+
+    cl = ngx_alloc_chain_link(r->pool);
```
dndx, line 1059: Maybe we do not have to reallocate a new chain link it seems? Simply overriding the old chain link's `buf` pointer should be enough.

src/ngx_http_lua_kong_module.c
dndx: Yes, the length need to be calculated based on the actual header size, instead of the original headers.

src/ngx_http_lua_kong_module.c
```diff
@@ -624,3 +713,562 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+
+            continue;
+        }
+
+        code = *(ngx_http_script_code_pt *) e.ip;
+        code((ngx_http_script_engine_t *) &e);
+
+        *e.pos++ = ':'; *e.pos++ = ' ';
+
+        if (ngx_strstr(e.pos - ngx_strlen("Host: "), "Host: ") != NULL &&
```
dndx, line 1149: `ngx_strlen("Host: ")` should be changed to `sizeof("Host: ") -1` to avoid runtime penalties.

Also, should we check for lowercased version as well? Maybe use `ngx_strcasestrn` instead.

src/ngx_http_lua_kong_module.c
```diff
@@ -624,3 +713,562 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+            r->headers_in.host->value.data != NULL) {
+            // keep the end of the Host header name position
+            orig_pos = e.pos;
+            // run the script
+            while (*(uintptr_t *) e.ip) {
+                code = *(ngx_http_script_code_pt *) e.ip;
+                code((ngx_http_script_engine_t *) &e);
```
dndx, line 1156: Question: do we really need this logic? If `proxy_set_header      Host               $upstream_host;` is set inside the `nginx.conf`, can we simply set `ngx.var.upstream_host = "new.host"` inside balancer phase for the script engine to take care of the rest?

If it is caused by weird caching script engine does, we can always clear the cache before running the length and content codes below. That way we can guarantee the length calculated is always the same as the string written and avoid the buffer mangling below.

dndx:
My recommendations:

Overall the change is going the right direction. However, the duplication of code from `ngx_http_proxy_module.c` is not ideal. Since the functions are almost identical to those of the `proxy_module`, maybe we should:

1. Introduce a patch file to Nginx to make `ngx_http_proxy_create_request` public so this module can call it.
2. Make any necessary change to `ngx_http_proxy_create_request` to make sure it can be safely called multiple times. This could includes adding additional parameters to the function. Since it is a private method in Nginx, nobody except the code inside `ngx_http_proxy_module.c` will be affected by it.
3. We should aim to make minimum changes to the `ngx_http_proxy_create_request` itself. If the `Host` header can be re-rendered from a variable, it should not be manipulated manually with string manipulation functions.

This way we should be able to get rid of majority of the copied code inside this PR, which makes the PR much more reviewable and maintainable, but most importantly, remain safe across Nginx updates.

dndx:
Superseded by https://github.com/openresty/lua-resty-core/pull/302.

PR #18, opened 2021-08-17
fix(grpc) allow setting authority on balancer phase

gszr:


dndx (approved):

dndx:
@gszr We should update the `README.md` file to include the new phase as well.

PR #19, opened 2021-08-31
feat(vars) add Lua API to use index to access variables

fffonion:


lib/resty/core/var.lua
```diff
@@ -0,0 +1,233 @@
+-- Copyright (C) Yichun Zhang (agentzh)
```
fffonion, line 1: TODO: make this a patch instead of a whole file

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,251 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+    }
+
+    v = ngx_array_push(&cmcf->variables);
+    if (v == NULL) {
+        *err = "no memory";
+        return NGX_ERROR;
+    }
+
+    v->name.len = name.len;
+    v->name.data = ngx_pnalloc(ngx_cycle->pool, name.len);
```
fffonion, line 760: if we limit the function to be called say init_* phases, will this pool be safe to use here?

t/var.lua
```diff
@@ -0,0 +1,229 @@
...
+    ngx_lua_ffi_var_get = C.ngx_stream_lua_ffi_var_get
+    ngx_lua_ffi_var_set = C.ngx_stream_lua_ffi_var_set
+end
+
+
+local value_ptr = ffi_new("unsigned char *[1]")
+local int_ptr = ffi_new("unsigned int [1]")
+local errmsg = base.get_errmsg_ptr()
+
+local function var_load_index(name)
```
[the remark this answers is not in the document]
fffonion: The index will be maintained by (patched) lua-resty-core and user won't notice a difference after this patch. Even we
return the actual index from `var_load_index`, it's more for debugging purpose, user are not allowed to get/set a
variable even with a index.
I think `ngx_http_get_variable_index` does a O(n) lookup and it could be expensive if we don't load it ahead of time.

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,223 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+    }
+
+    cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+    if (index >= cmcf->variables.nelts) {
+       *err = "variable index out of range";
+        return NGX_ERROR;
+    }
+
+    v = ((ngx_http_variable_t*) cmcf->variables.elts) + index;
```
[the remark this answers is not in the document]
fffonion: `ngx_http_get_indexed_variable` returns the `ngx_http_variable_value_t` while setting the variable requires
the `ngx_http_variable_t` since the set handler needs to be called. [ngx_http_get_indexed_variable](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_variables.c#L605) actually also first
get the `ngx_http_variable_t` then invokes `get_handler` to return the value similar to the flow we did here.

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,247 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+            {
+                continue;
+            }
+
+            *index = i;
+            return NGX_OK;
+        }
+    }
+
+    v = ngx_array_push(&cmcf->variables);
```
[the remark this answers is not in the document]
fffonion: Yeah I agree it might be a bad idea if we allow user to call this after config phase, if it's during request then on L760
we may end up allocating the variable in the request memory pool (correct me if i'm wrong) and that's bad.
How do you feel if we restrict the API to has to be called in `init_by_lua` as it's technically still in config phase?
The reason I didn't manage to use `ngx_http_get_variable_index` here is it takes a `ngx_conf_t` as parameter and I think
that make it only possible to be called inside a directive 🤔

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,223 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
 
     return ctx->upstream_ssl_verify;
 }
+# endif
+
+int
+ngx_http_lua_kong_ffi_var_load_index(u_char *name_data,
```
fffonion, line 689: change to a directive and a single FFI function to get all indexes from known variables.

lib/resty/core/var.lua
fffonion: go with the monkeypatch approach, resolving this.

README.md
```diff
@@ -297,6 +329,23 @@ called from the current session.
...
+
+**context:** *any*
+
+**subsystems:** *http*
+
+Replace the get/set method of ngx.var to use index. Variables that are not indexed
+will fallback to use the original get/set method.
+
+To index an variable before access, user may use [lua_kong_load_var_index](#lua_kong_load_var_index)
+directive.
```
dndx, line 344:
```suggestion
Indexed variable access is a faster way of accessing Nginx variables for OpenResty.
This method patches the metatable of `ngx.var` to enable index access to variables
that supports it. It should be called once in the `init` phase which will be effective for
all subsequent `ngx.var` uses.

For variables that does not have indexed access, the slower hash based lookup will
be used instead (this is the OpenResty default behavior).

To ensure a variable can be accessed using index, you can use the [lua_kong_load_var_index](#lua_kong_load_var_index)
directive.
```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+local function load_indexes(count)
+    count = count or 100
+    local names_buf = ffi_new("ngx_str_t *[?]", count)
+    local count_ptr = ffi_new("unsigned int [1]")
+    count_ptr[0] = count
+
+    local rc = ngx_lua_kong_ffi_var_load_indexes(names_buf,
+                                                 count_ptr, errmsg)
+
+    if rc == 0 then -- NGX_OK
```
dndx, line 55: We should use `ngx.OK`, `ngx.ERR` (after module caching), etc for those core constants. Same below.

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+        count = tonumber(count_ptr[0])
```
dndx, line 56: `tonumber` should be cached as well. Maybe we should lint on this file using `lualint` just to catch all of those.

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+        for i = 0, count do
+            local name = ffi_str(names_buf[i].data, names_buf[i].len)
+            ngx.log(ngx.ERR, name, " => ", i)
```
dndx, line 59: We should remove this.

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+        value_len = 0
+    else
+        if type(value) ~= 'string' then
+            value = tostring(value)
+        end
+        value_len = #value
+    end
+
+    local rc = ngx_lua_kong_ffi_var_set_by_index(r, index, value,
+                                                      value_len, errmsg)
```
dndx, line 111:
```suggestion
                                                 value_len, errmsg)
```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+
+    -- ngx.log(ngx.WARN, "rc = ", rc)
```
dndx, line 113: Should remove this as well.

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,161 @@
...
+    if rc == 0 then -- NGX_OK
+        return
+    end
+
+    if rc == -1 then  -- NGX_ERROR
+        error(ffi_str(errmsg[0]), 2)
+    end
+end
+
+local function patch_metatable()
```
dndx, line 124: As mentioned above, we should:

1. Check to make sure the phase is `init`.
2. Make sure `patch_metatable` can be called only once, subsequent calls should throw an `error` to warn the user.

src/ngx_http_lua_kong_module.c
```diff
@@ -48,11 +49,23 @@ static ngx_http_module_t ngx_http_lua_kong_module_ctx = {
     NULL                                     /* merge location configuration */
 };
 
+static ngx_command_t ngx_http_lua_kong_cmds[] = {
+
+    { ngx_string("lua_kong_load_var_index"),
+      NGX_HTTP_MAIN_CONF|NGX_CONF_TAKE1,
+      ngx_http_lua_kong_load_var_index,
+      NGX_HTTP_LOC_CONF_OFFSET,
```
dndx, line 57:
```suggestion
      0,
```

Since there is no `conf` associated with this directive, it does not need a valid offset in this case.

src/ngx_http_lua_kong_module.c
```diff
@@ -48,11 +49,23 @@ static ngx_http_module_t ngx_http_lua_kong_module_ctx = {
     NULL                                     /* merge location configuration */
 };
 
+static ngx_command_t ngx_http_lua_kong_cmds[] = {
```
dndx, line 52:
```suggestion
static ngx_command_t ngx_http_lua_kong_commands[] = {
```

This is consistent with Nginx's module naming conventions.

src/ngx_http_lua_kong_module.c
```diff
@@ -48,11 +49,23 @@ static ngx_http_module_t ngx_http_lua_kong_module_ctx = {
...
+
+    ngx_null_command
+};
+
 
 ngx_module_t ngx_http_lua_kong_module = {
     NGX_MODULE_V1,
     &ngx_http_lua_kong_module_ctx,     /* module context */
-    NULL,                              /* module directives */
+    ngx_http_lua_kong_cmds,            /* module directives */
```
dndx, line 68:
```suggestion
    ngx_http_lua_kong_commands,        /* module directives */
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
 
     return ctx->upstream_ssl_verify;
 }
+# endif
+
+char *
+ngx_http_lua_kong_load_var_index(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+
```
dndx, line 691:
```suggestion
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+                           "invalid variable name size \"%V\"",
+                           &value[1]);
+        return NGX_CONF_ERROR;
+    }
+
+    index = ngx_http_get_variable_index(cf, &value[1]);
+
+    if (index == NGX_ERROR) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "undefined variable \"%V\"",
```
dndx, line 708: I think `NGX_ERROR` returned basically means memory allocation failures, we should change this error message to be better reflecting that.
```suggestion
                           "unable to mark variable \"%V\" as indexed: no memory",
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+char *
+ngx_http_lua_kong_load_var_index(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+
+    ngx_str_t                     *value;
+    ngx_int_t                      index;
+
+    value = cf->args->elts;
+
+    if (value[1].len == 0) {
```
dndx, line 697: If we make the argument starts with `$`, then the following logic can be copied directly from `set` directive:

http://lxr.nginx.org/source/xref/nginx/src/http/modules/ngx_http_rewrite_module.c?r=7606%3Adb8df9cd84c8#908

In that case, the `value[1].len == 0` check is not necessary. L908-L915 is probably enough here.

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+{
+    ngx_uint_t                  i;
+    ngx_http_variable_t        *v;
+    ngx_http_core_main_conf_t  *cmcf;
+
+    cmcf = ngx_http_cycle_get_module_main_conf(ngx_cycle, ngx_http_core_module);
+
+    v = cmcf->variables.elts;
+
+    assert(v != NULL);
```
dndx, line 728: We can use `ngx_http_lua_assert` here, and it will only be enabled for debug builds (Nginx and OpenResty's release build contains no asserts).

See:

https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_common.h#L39-L44
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_cache.c#L62

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+
+    for (i = 0; i < cmcf->variables.nelts && i < *count; i++) {
```
dndx, line 730: A better logic is the following:

When `ngx_http_lua_kong_ffi_var_load_indexes` is called the first time, `names` will be `NULL`, this function returns `cmcf->variables.nelts` to let the caller know how big an array it needs.

Then `names` will be allocated according to the return value. In this case `count` won't even be needed at all (this is a common pattern in OpenResty).

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+        *err = "API disabled in the current context";
+        return NGX_ERROR;
+    }
+
+    cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+    if (index >= cmcf->variables.nelts) {
+       *err = "variable index out of range";
+        return NGX_ERROR;
+    }
```
dndx, line 790: We probably don't need this. (or change to `ngx_http_lua_assert` should be enough).

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+
+    v = ((ngx_http_variable_t*) cmcf->variables.elts) + index;
```
dndx, line 792:
```suggestion
    v = ((ngx_http_variable_t *) cmcf->variables.elts) + index;
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+     * openresty/lua-nginx-module/blob/master/src/ngx_http_lua_variable.c
+     */
+
+    if (!(v->flags & NGX_HTTP_VAR_CHANGEABLE)) {
+        *err = "variable not changeable";
+        return NGX_ERROR;
+    }
+
+    if (v->set_handler) {
+
```
dndx, line 805:
```suggestion
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +683,197 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+            vv->valid = 1;
+            vv->not_found = 0;
+            vv->no_cacheable = 0;
+
+            vv->data = value;
+            vv->len = value_len;
+        }
+
+        v->set_handler(r, vv, v->data);
+        return NGX_OK;
```
dndx, line 841:
```suggestion

        return NGX_OK;
```

src/ngx_http_lua_kong_module.c
fffonion: done

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+
+    local count = ngx_lua_kong_ffi_var_load_indexes(nil)
+    count = tonumber(count)
+
+    local names_buf = ffi_new("ngx_str_t *[?]", count)
+
+    local rc = ngx_lua_kong_ffi_var_load_indexes(names_buf)
+
+    if rc == NGX_OK then
+        for i = 0, count-1 do
```
dndx, line 63:
```suggestion
        for i = 0, count - 1 do
```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+
+local function var_set_by_index(index, value)
+    local r = get_request()
+    if not r then
+        error("no request found")
+    end
+
+    local value_len
+    if value == nil then
+        value_len = 0
```
dndx, line 103:
```suggestion
        value_len = 0

```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+    else
+        if type(value) ~= 'string' then
+            value = tostring(value)
+        end
```
dndx, line 107:
```suggestion
        end

```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+                                                 value_len, errmsg)
+
+    if rc == NGX_OK then
+        return
+    end
+
+    if rc == NGX_ERROR then
+        error(ffi_str(errmsg[0]), 2)
+    end
+end
```
dndx, line 121:
```suggestion
    assert(rc == NGX_ERROR)
    error(ffi_str(errmsg[0]), 2)
end
```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+local function patch_metatable()
+    if ngx.get_phase() ~= "init" then
+        error("patch_metatable can only be called in init phase")
+    end
+
+    if metatable_patched then
+        error("patch_metatable should only be called once")
+    end
+
+    patch_metatable = true
```
dndx, line 132:
```suggestion
    metatable_patched = true
```

lualib/resty/kong/var.lua
```diff
@@ -0,0 +1,168 @@
...
+        return ffi_str(value_ptr[0], value_len[0])
+    end
+
+    if rc == NGX_DECLINED then
+        return nil
+    end
+
+    if rc == NGX_ERROR then
+        error(ffi_str(errmsg[0]), 2)
+    end
```
dndx, line 92:
```suggestion
    assert(rc == NGX_ERROR)
    error(ffi_str(errmsg[0]), 2)
```

src/ngx_http_lua_kong_module.c
```diff
@@ -29,6 +29,14 @@ static int ngx_http_lua_kong_ssl_no_session_cache_flag_index = -1;
 static int
 ngx_http_lua_kong_verify_callback(int ok, X509_STORE_CTX *x509_store);
 #endif
+
+#ifdef NGX_LUA_USE_ASSERT
+#include <assert.h>
+#   define ngx_http_lua_assert(a)  assert(a)
```
dndx, line 35:
```suggestion
#   define ngx_http_lua_kong_assert(a)  assert(a)
```

Avoid using the same name as OpenResty's header, otherwise if we `include` it in the future it will cause conflicts.

src/ngx_http_lua_kong_module.c
```diff
@@ -29,6 +29,14 @@ static int ngx_http_lua_kong_ssl_no_session_cache_flag_index = -1;
...
+#else
+#   define ngx_http_lua_assert(a)
```
fffonion, line 37:
```suggestion
#   define ngx_http_lua_kong_assert(a)
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +690,198 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+    }
+
+    if ((r)->connection->fd == (ngx_socket_t) -1) {
+        *err = "API disabled in the current context";
+        return NGX_ERROR;
+    }
+
+    cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+    ngx_http_lua_assert(index < cmcf->variables.nelts);
```
fffonion, line 798:
```suggestion
    ngx_http_lua_kong_assert(index < cmcf->variables.nelts);
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +690,198 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+    cmcf = ngx_http_cycle_get_module_main_conf(ngx_cycle, ngx_http_core_module);
+
+    /* return required size only */
+    if (names == NULL) {
+        return cmcf->variables.nelts;
+    }
+
+    v = cmcf->variables.elts;
+
+    ngx_http_lua_assert(v != NULL);
```
fffonion, line 740:
```suggestion
    ngx_http_lua_kong_assert(v != NULL);
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +690,198 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+
+    for (i = 0; i < cmcf->variables.nelts; i++) {
+        ngx_http_lua_assert(v[i].index == i);
```
fffonion, line 743:
```suggestion
        ngx_http_lua_kong_assert(v[i].index == i);
```

src/ngx_http_lua_kong_module.c
```diff
@@ -670,5 +690,198 @@ ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
...
+            vv->data = value;
+            vv->len = value_len;
+        }
+
+        v->set_handler(r, vv, v->data);
+
+        return NGX_OK;
+    }
+
+    ngx_http_lua_assert(v->flags & NGX_HTTP_VAR_INDEXED);
```
fffonion, line 852:
```suggestion
    ngx_http_lua_kong_assert(v->flags & NGX_HTTP_VAR_INDEXED);
```

dndx (approved):
Valgrind test: no issues found
Leak test: no issue found

LGTM. Nice work @fffonion!

PR #21, opened 2021-12-15
chore(*) add couple of functions for better macOS+M1 support

bungle:
### Summary

There is a LuaJIT bug that is not yet resolved and it causes issues with function calls on macOS+M1. Apple has decided to
break ABI compatibility with ARM64, and thus here we workaround the issue described here:

https://github.com/LuaJIT/LuaJIT/issues/205

Proper fix is to fix LuaJIT, but as that issue has been open since 2016, I thought it would probably be faster to fix this with couple of forward declarations.

This also needs changes in `resty.core.shdict` and `resty.core.response` to take advantage of these new functions. New functions just take struct in and forward the call on C-side to original functions. The issue arises with macOS+M1 if functions have around 8+ arguments. I will link this issue to fixes to above mentioned files, when I decide what is the best way to go with those.

Alternatively we could do this as patch to OpenResty in `kong-build-tools`, but I felt like this module is good home for it.

This commit is related to:
https://github.com/Kong/kong/pull/8202
https://github.com/Kong/kong-build-tools/pull/427

### DIFF
--- a/src/ngx_http_lua_kong_module.c
+++ b/src/ngx_http_lua_kong_module.c
@@ -885,3 +885,40 @@ nomem:
     *err = "no memory";
     return NGX_ERROR;
 }
+
+// macOS with M1 fixes, see: https://github.com/LuaJIT/LuaJIT/issues/205
+
+int
+ngx_http_lua_ffi_shdict_get_m1(ngx_shdict_get_t *s)
+{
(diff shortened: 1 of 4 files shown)


bungle:
One note: currently it does not have support for `stream module` functions.

kikito (approved):

PR #24, opened 2021-12-25
fix(tls) remove depth param in ffi_set_upstream_ssl_verify

chronolaw:
I think it is an error in C declaration of ffi_set_upstream_ssl_verify.

There is a redundancy param depth, 
and in lua-resty code, the declaration is 

int ngx_http_lua_kong_ffi_set_upstream_ssl_verify(ngx_http_request_t *r,
        int verify);


fffonion (approved):

fffonion:
Good catch!

PR #28, opened 2022-02-16
chore(*) seperate source code

chronolaw:
Following ngx_lua's style, seperate original ngx_http_lua_kong_module.c into several *.c.


src/ngx_http_lua_kong_ssl.c
```diff
@@ -0,0 +1,592 @@
...
+ngx_http_lua_kong_cleanup_trusted_store(ngx_http_lua_kong_ctx_t *ctx)
+{
+    if (ctx->upstream_trusted_store != NULL) {
+        X509_STORE_free(ctx->upstream_trusted_store);
+    }
+}
+
+
+static void
+ngx_http_lua_kong_cleanup(void *data)
```
fffonion, line 409: let's add "namespace" for this function, as it's only used inside the ssl part. how about `ngx_http_lua_kong_ssl_cleanup`, so we can avoid collision of symbols.

src/ngx_http_lua_kong_ssl.c
```diff
@@ -0,0 +1,592 @@
...
+    ngx_http_lua_kong_ctx_t     *ctx = data;
+
+    ngx_http_lua_kong_cleanup_cert_and_key(ctx);
+
+    ngx_http_lua_kong_cleanup_trusted_store(ctx);
+}
+
+
+ngx_http_lua_kong_ctx_t *
+ngx_http_lua_kong_get_module_ctx(ngx_http_request_t *r)
```
fffonion, line 420: same as above comment, as this function calls `ngx_http_lua_kong_cleanup`, let's also namespace it for now and call it like `ngx_http_lua_kong_ssl_get_module_ctx`. if there's other module want to get_module_ctx, let's abstract it out later.

src/ngx_http_lua_kong_ssl.c
chronolaw: It is used also in grpc, so we must keep this name.

src/ngx_http_lua_kong_ssl.c
chronolaw: moved into module.c.

fffonion (approved):
feel free to merge after CI passed

fffonion (approved):

PR #37, opened 2022-03-15
tests(github) add boringssl

fffonion:


src/ngx_http_lua_kong_ssl.c
```diff
@@ -521,7 +521,15 @@ ngx_http_lua_kong_ffi_set_upstream_ssl_verify_depth(ngx_http_request_t *r,
     }
 
     ctx->upstream_ssl_verify_depth_set = 1;
+#ifdef OPENSSL_IS_BORINGSSL
```
fffonion, line 524: just to see if this fix the CI, the correct way to fix this is to patch nginx, as all parts involves ssl verify changes its behaviour

PR #38, opened 2022-03-21
style(*) code clean for m1

chronolaw:
Code clean for nginx style.

### DIFF
--- a/src/ngx_http_lua_kong_apple_m1.c
+++ b/src/ngx_http_lua_kong_apple_m1.c
@@ -20,11 +20,6 @@
 
 // macOS with M1 fixes, see: https://github.com/LuaJIT/LuaJIT/issues/205
 
-int ngx_http_lua_ffi_shdict_get(ngx_shm_zone_t *zone, const unsigned char *key,
-    size_t key_len, int *value_type, unsigned char **str_value_buf,
-    size_t *str_value_len, double *num_value, int *user_flags,
-    int get_stale, int *is_stale, char **errmsg);
-
 typedef struct {
     ngx_shm_zone_t *zone;
     const unsigned char *key;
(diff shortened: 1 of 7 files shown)


src/ngx_http_lua_kong_apple_m1.c
```diff
@@ -20,11 +20,6 @@
 
 // macOS with M1 fixes, see: https://github.com/LuaJIT/LuaJIT/issues/205
```
dndx, line 21: Also should remove this or change the style to C89.

src/ngx_http_lua_kong_apple_m1.c
```diff
@@ -39,21 +34,6 @@ typedef struct {
     char **errmsg;
 } ngx_shdict_get_t;
```
dndx, line 35: Let's move these into a separate header file.

dndx (commented):
Basically LGTM, we should probably change the filename as well. Maybe `src/ngx_http_lua_kong_macos_shdict.c` is better.

src/ngx_http_lua_kong_apple_m1.c
chronolaw: thanks

src/ngx_http_lua_kong_apple_m1.c
dndx: Also, we should probably rename this struct to something like `ngx_http_lua_kong_shdict_get_t` or something similar. `ngx_shdict_get_t` seems to be a bit odd.

src/ngx_http_lua_kong_macos_arm64.h
```diff
@@ -0,0 +1,116 @@
...
+#ifndef _NGX_HTTP_LUA_KONG_MACOS_ARM64_H_INCLUDED_
+#define _NGX_HTTP_LUA_KONG_MACOS_ARM64_H_INCLUDED_
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_http.h>
+
+
+/* macOS with Apple Silicon fixes, see: https://github.com/LuaJIT/LuaJIT/issues/205 */
```
dndx, line 27:
```suggestion
/*
 * macOS with Apple Silicon fixes,
 * see: https://github.com/LuaJIT/LuaJIT/issues/205
 */
```

Just minor Nginx style correction.

bungle (approved):

PR #41, opened 2022-05-18
feat(*) a new static tag feature

chronolaw:
I noticed that we use some vars as semantic tag in nginx config file, 
such as `$kong_proxy_mode`.

It works well but I don't think it is the proper usage of nginx vars.

So I add a new directive `lua_kong_set_static_tag`, which will store the static string in the module's loc_conf_t,
and we can access the value very fast without the cost of var hash searching.

I also did a simple performance test, get the value for 5000*1000 times by `ngx.var` and `tag.get`,
the result is:

0.21599984169006
0.005000114440918

You can see `tag.get` is extremely faster  than `ngx.var`.

If you have interest in this feature I will improve the code, please give me some suggestions.

### DIFF
--- a/t/007-static-tag.t
+++ b/t/007-static-tag.t
@@ -62,3 +62,37 @@ value:nil
 
 
 
+=== TEST 3: set tag for internal location block
+--- http_config
+    lua_package_path "../lua-resty-core/lib/?.lua;lualib/?.lua;;";
+--- config
+    location = /test {
+        lua_kong_set_static_tag "test-tag";
+        content_by_lua_block {
+            local tag = require "resty.kong.tag"
+            ngx.say("value:", tag.get())
+
+            local res = ngx.location.capture("/inner")
+            ngx.say(res.body)
+        }
+    }
+    location = /inner {
+        internal;
+        lua_kong_set_static_tag "inner-tag";
+        content_by_lua_block {
+            local tag = require "resty.kong.tag"
+            ngx.print("value:", tag.get())
+        }
+    }
+--- request
+GET /test
+--- response_body
+value:test-tag
+value:inner-tag
+--- no_error_log
+[error]
+[crit]
+[alert]
+
+
+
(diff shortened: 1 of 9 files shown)


flrgh:
I like this idea. In my own OpenResty projects this need to associate metadata with a `location` or `server` config has come up several times.

Given this feature, I think the next logical step would be adding a Lua API that uses the tag as a lookup key so that more metadata can be used in Lua-land:

```nginx
location / {
  set_static_tag "root";
}

location @other {
  set_static_tag "other";
}
```

```lua
-- kong.meta.lua

local meta = {
  root = {
    attr = 123,
  },
  other = {
    attr = 456,
  }
}

return {
  get_location_meta = function()
    return meta[get_static_location_tag_name()]
  end,
}
```

I think it would be helpful to make this applicable in the `server` context as well, with a separate namespace:

```nginx
server {
  listen 80;
  set_static_tag "my server tag";
  
  location / {
    set_static_tag "my location tag";

    content_by_lua_block {
      ngx.print(get_server_tag()) -- "my server tag"
      ngx.print(get_location_tag()) -- "my location tag"
    }

    location /nested/inherit {
      content_by_lua_block {
        ngx.print(get_server_tag()) -- "my server tag"
        ngx.print(get_location_tag()) -- "my location tag"
      }
    }

    location /nested/override {
      set_static_tag "my nested location tag";
      content_by_lua_block {
        ngx.print(get_server_tag()) -- "my server tag"
        ngx.print(get_location_tag()) -- "my nested location tag"
      }
    }
  }
}
```

chronolaw:
Thanks for your opinions. I will consider this, maybe implement it soon.

chronolaw:
> this is a very good idea! what will be behaviour be to access tag in an internal location block and in sub_request?

In short words, it is still a cheaper way to replace ngx.var usage, but its target is more special.

t/007-static-tag.t
```diff
@@ -0,0 +1,124 @@
...
+
+
+=== TEST 4: set tag for nested location block
+--- http_config
+    lua_package_path "../lua-resty-core/lib/?.lua;lualib/?.lua;;";
+--- config
+    location /test {
+        lua_kong_set_static_tag "test-tag";
+
+        location /test/nested {
```
fffonion, line 106: i didn't know nginx support this syntax till today

fffonion (approved):

PR #42, opened 2022-05-24
chore(*) remove m1 ffi functions, fixed in upstream

bungle:
This got fixed on `OpenResty 1.21.4.1`.


chronolaw:
I think OpenResty 1.21.4.1 does not include the apple m1 patches now, maybe it will happen in 1.21.4.2.

PR #43, opened 2022-08-31
feat(var) use ffi to get some request vars

chronolaw:
I collected some var in kong's hot path, then try to use ffi to access them.

Here is the list:
* $https
* $scheme
* $is_args
* $args
* $request_uri
* $server_port


chronolaw:
I did some simple test, iteration 10*5000*1000 times:

ngx.var.https: 0.93099999427795
req.is_https : 0.047000169754028

ngx.var.scheme: 1.1800000667572
req.scheme       : 0.046999931335449

ngx.var.request_uri: 1.4860000610352
req.request_uri : 0.34200000762939

chronolaw:
Close. We can reopen it if we need it in the future.

PR #46, opened 2022-10-19
feat(tls) add new API `tls.set_client_ca_list(ca_list)`

catbro666:
This API adds the CA DN list to the underlying SSL structure, which will be sent in the Certificate Request message.

The client then can use this DN information to filter the certs and chooses an appropriate cert issued by a CA in the list.

FTI-4430

### DIFF
--- a/lualib/resty/kong/tls.lua
+++ b/lualib/resty/kong/tls.lua
@@ -27,6 +27,9 @@ if ngx.config.subsystem == "http" then
     int ngx_http_lua_kong_ffi_get_full_client_certificate_chain(
         ngx_http_request_t *r, char *buf, size_t *buf_len);
     const char *ngx_http_lua_kong_ffi_disable_session_reuse(ngx_http_request_t *r);
+    // STACK_OF(X509_NAME)
+    const char *ngx_http_lua_kong_ffi_set_client_ca_list(ngx_http_request_t *r,
+        OPENSSL_STACK *name_list);
     int ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
         void *_chain, void *_key);
     int ngx_http_lua_kong_ffi_set_upstream_ssl_trusted_store(ngx_http_request_t *r,
@@ -91,6 +94,22 @@ if ngx.config.subsystem == "http" then
         local r = get_request()
 
         local errmsg = C.ngx_http_lua_kong_ffi_disable_session_reuse(r)
+        if errmsg == nil then
+            return true
+        end
+
+        return nil, ffi_string(errmsg)
+    end
+
+
+    function _M.set_client_ca_list(name_list)
+        if get_phase() ~= 'ssl_cert' then
+            error("API disabled in the current context")
+        end
+
+        local r = get_request()
+
+        local errmsg = C.ngx_http_lua_kong_ffi_set_client_ca_list(r, name_list)
         if errmsg == nil then
             return true
         end
--- a/lualib/resty/kong/tls.lua
+++ b/lualib/resty/kong/tls.lua
@@ -29,7 +29,7 @@ if ngx.config.subsystem == "http" then
     const char *ngx_http_lua_kong_ffi_disable_session_reuse(ngx_http_request_t *r);
     // STACK_OF(X509_NAME)
     const char *ngx_http_lua_kong_ffi_set_client_ca_list(ngx_http_request_t *r,
-        OPENSSL_STACK *name_list);
+        void *name_list);
     int ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
         void *_chain, void *_key);
     int ngx_http_lua_kong_ffi_set_upstream_ssl_trusted_store(ngx_http_request_t *r,
--- a/lualib/resty/kong/tls.lua
+++ b/lualib/resty/kong/tls.lua
@@ -27,9 +27,9 @@ if ngx.config.subsystem == "http" then
     int ngx_http_lua_kong_ffi_get_full_client_certificate_chain(
         ngx_http_request_t *r, char *buf, size_t *buf_len);
     const char *ngx_http_lua_kong_ffi_disable_session_reuse(ngx_http_request_t *r);
-    // STACK_OF(X509_NAME)
+    // STACK_OF(X509)
     const char *ngx_http_lua_kong_ffi_set_client_ca_list(ngx_http_request_t *r,
-        void *name_list);
+        void *ca_list);
     int ngx_http_lua_kong_ffi_set_upstream_client_cert_and_key(ngx_http_request_t *r,
         void *_chain, void *_key);
     int ngx_http_lua_kong_ffi_set_upstream_ssl_trusted_store(ngx_http_request_t *r,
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_kong_ssl.c
```diff
@@ -179,6 +179,42 @@ ngx_http_lua_kong_ffi_request_client_certificate(ngx_http_request_t *r)
...
+    ngx_ssl_conn_t      *sc;
+
+    if (c->ssl == NULL) {
+        return "server does not have TLS enabled";
+    }
+
+    sc = c->ssl->connection;
+
+
+    SSL_set_client_CA_list(sc, name_list);
```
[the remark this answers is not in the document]
catbro666: I also noticed this and deep copied the `name_list` in the outer layer. I think you're right, it's better to do this within the interface. BTW, `X509_chain_up_ref` works on `STACK_OF(X509) *`, and `X509_NAME` has no reference counter, so we can only deep copy it.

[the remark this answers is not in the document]
fffonion: ah yes, i misread the signature, it's a `STACK_OF(X509_NAME)`. however, there isn't a convient function in openssl that creates a `STACK_OF(X509_NAME)`, maybe possible with some development with `resty.openssl.stack`, but that makes this API not immediately available.

so i would suggest a bit change to the API, instead of accepting a `STACK_OF(X509_NAME)`, let's pass a `STACK_OF(X509)` (which is then a `resty.openssl.x509.chain`.

in `ngx_http_lua_kong_ffi_set_client_ca_list`, we will 1) create a stack_of(x509_name), 2) iterate over the x509 stack, and generate X509_NAME from each of the cert, 3) add to the stack, 4) call SSL_set_client_CA_list. don't need to free the stack or the x509_name, it will be freed together with the `SSL` instance.

lualib/resty/kong/tls.lua
```diff
@@ -99,6 +102,22 @@ if ngx.config.subsystem == "http" then
     end
 
 
+    function _M.set_client_ca_list(name_list)
+        if get_phase() ~= 'ssl_cert' then
+            error("API disabled in the current context")
+        end
+
+        local r = get_request()
```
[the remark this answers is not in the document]
fffonion: this is more for protective code, and is a good pratice rather than to fix a bug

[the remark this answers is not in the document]
catbro666: gotcha

src/ngx_http_lua_kong_ssl.c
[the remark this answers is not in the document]
catbro666: Good suggestion! Taking `STACK_OF(X509) *` as a parameter does make the lua code simpler and thus alleviates the burden of users.

lualib/resty/kong/tls.lua
```diff
@@ -27,6 +27,9 @@ if ngx.config.subsystem == "http" then
     int ngx_http_lua_kong_ffi_get_full_client_certificate_chain(
         ngx_http_request_t *r, char *buf, size_t *buf_len);
     const char *ngx_http_lua_kong_ffi_disable_session_reuse(ngx_http_request_t *r);
+    // STACK_OF(X509)
```
chronolaw, line 30: This line is unnecessary.

catbro666: It's a hint of the type of the opaque pointer.

catbro666:
> 

Added.

t/001-tls.t
```diff
@@ -521,3 +521,206 @@ nil, connection is not TLS or TLS support for Nginx not enabled
...
+
+--- error_log
+ssl cert by lua is running!
+ssl cert by lua complete!
+
+--- no_error_log
+[error]
+[alert]
+[warn]
+[crit]
```
attenuation, line 726: add test for `ssl.parse_pem_cert` function to get `STACK_OF(X509) *`?

catbro666: I think it should be a test of `ngx.ssl` itself. We only focus on the APIs of our own module.

fffonion: i agree having such test will be good, as `ssl.parse_pem_cert` doesn't guaranteed to always work with our library, it's designed to be consumed
by another upstream API, so it could change internally without our notice.

lualib/resty/kong/tls.lua
chronolaw: I checked the source code of lua-resty-core, there is no comments like this.

I think we need not care about the pointer type in Lua land.

lualib/resty/kong/tls.lua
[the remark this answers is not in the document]
chronolaw: I checked the other function's code, like `request_client_certificate` `disable_session_reuse`, all of them have no request check.

And there is a comment:
```
        -- no need to check if r is nil as phase check above
        -- already ensured it
```
So I think it is OK.

lualib/resty/kong/tls.lua
catbro666: Actually, I copied this style from `lua-resty-openssl`😂. For example, [here](https://github.com/fffonion/lua-resty-openssl/blob/master/lib/resty/openssl/include/x509_vfy.lua#L17&L31). If this is not preferred, we can remove it.

lualib/resty/kong/tls.lua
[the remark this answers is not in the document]
catbro666: Yes, logically no need to check, but as a good practice, it's better to check. Another tradeoff.

t/001-tls.t
catbro666: Okay, I will add the test.

src/ngx_http_lua_kong_ssl.c
```diff
@@ -179,6 +179,51 @@ ngx_http_lua_kong_ffi_request_client_certificate(ngx_http_request_t *r)
...
+        /* will call X509_NAME_dup() internally */
+        if (SSL_add_client_CA(sc, ca) == 0) {
+            return "unable to add the CA name to the list";
+        }
+    }
+
+    return NULL;
+
+#else
+    return "TLS support is not enabled in Nginx build"
```
chronolaw, line 222: need a ';' at the end of line.

src/ngx_http_lua_kong_ssl.c
```diff
@@ -179,6 +179,51 @@ ngx_http_lua_kong_ffi_request_client_certificate(ngx_http_request_t *r)
...
+    X509                *ca;
+    int                  i;
+
+    if (c->ssl == NULL) {
+        return "server does not have TLS enabled";
+    }
+
+    sc = c->ssl->connection;
+
+
```
chronolaw, line 209: keep only one blank line?

lualib/resty/kong/tls.lua
```diff
@@ -99,6 +101,25 @@ if ngx.config.subsystem == "http" then
     end
 
 
+    function _M.set_client_ca_list(ca_list)
+        if get_phase() ~= 'ssl_cert' then
+            error("API disabled in the current context")
+        end
+
+        local r = get_request()
+        if not r then
```
chronolaw, line 110: I know we already have the discussion about checking `r`,
but the other functions still don't do the same thing, 
Should we change the others to keep consistent？

fffonion: my opinion is to add this check to existing functions as well, although it's indeed redundent
given our phase check. but it gives us extra protection because this a private API from openresty,
it doesn't guaranteeded to be unchanged in the future.

chronolaw: I think we have gotten the agreement. Let's do it in another PR, keep this PR clean.

src/ngx_http_lua_kong_ssl.c
catbro666: Several other functions also need to add it, I only changed this one to keep this PR clean.

fffonion (approved):

PR #48, opened 2022-10-20
docs(readme): doc for lua_kong_set_static_tag

chronolaw:
Add directive and API description for PR #41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,8 @@ Return the tag value which is set by [`lua_kong_set_static_tag`](#lua_kong_set_s
 If there is no tag in `location`(http subsystems) or `server`(stream subsystems) block,
 it will return `nil`.
 
+[Back to TOC](#table-of-contents)
+
 
 License
 =======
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -393,6 +406,21 @@ directive.
 
 [Back to TOC](#table-of-contents)
 
+resty.kong.tag.get
+----------------------------------
+**syntax:** *resty.kong.tag.get()*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, log_by_lua&#42;*
```
flrgh, line 413: Is it usable in `header_filter` and `body_filter`?

chronolaw: I think it can work in these phases, let me add some test cases.

flrgh (approved):

README.md
```diff
@@ -121,6 +123,17 @@ indexed variable access.
...
 
+lua\_kong\_set\_static\_tag
+-------------------------------------------
+**syntax:** *lua_kong_set_static_tag value;*
+
+**context:** *location(http subsystem)* *server(stream subsystem)*
+
+Add a static tag string for Nginx's `location`(http subsystem) or `server`(stream subsystem) block,
+which can be accessed in Lua land by [`resty.kong.tag.get`](#restykongtagget).
+
```
fffonion, line 134: Let's decribe the scope of this tag, it can just reflect the behaviour we observed in tests (derive from server block to location, isolate beteen locations, works in nested location, doesn't work in sub query etc)

chronolaw: Add some description.

Another thing, this directive only works in `http/location` or `stream/server`, we can not define tags in `http/server` now. so no deriving from server block.

fffonion (approved):

PR #50, opened 2022-11-02
feat(stream) add upstream ssl ffi function and stream client ssl ffi function

attenuation:
1. extend stream module upstream ssl client function to support dynamic  upstream ssl cert
2. extend stream tls request client_certificate_chain


fffonion:
might consider make a common helper function to reduce copy paste. although `r` is different, `c`  and `ssl` is the same struct that shared between http and stream.

attenuation:
https://github.com/Kong/lua-kong-nginx-module/commit/befcfb0da8a91c479cca74ece45229b44735a526 this pr will be revert because openresty official API support this feature, wanting for this I will rebase to master.

src/ngx_http_lua_kong_common.h
```diff
@@ -22,17 +22,11 @@
...
 typedef struct {
-    STACK_OF(X509)     *upstream_client_certificate_chain;
-    EVP_PKEY           *upstream_client_private_key;
-    X509_STORE         *upstream_trusted_store;
-    ngx_uint_t          upstream_ssl_verify_depth;
-    ngx_str_t           grpc_authority;
-    unsigned            upstream_ssl_verify:1;
-    unsigned            upstream_ssl_verify_set:1;
-    unsigned            upstream_ssl_verify_depth_set:1;
+    ngx_lua_kong_ssl_ctx_t  *ssl_ctx;
```
chronolaw, line 28: `*` should follow nginx's align style.

catbro666:
How about adjusting the codes under `stream/src` into `src/stream`, i think it’s more proper there.

attenuation:
@catbro666     I think it's not needed because many Nginx modules use this file struct in the code base.

stream/src/ngx_stream_lua_kong_module.h
```diff
@@ -5,10 +5,12 @@
...
 #include <ngx_core.h>
 #include <ngx_stream.h>
-
+#include <ngx_stream_lua_api.h>
+#include "../../src/ssl/ngx_lua_kong_ssl.h"
 
 typedef struct {
-    ngx_uint_t              proxy_ssl_disable; /* unsigned proxy_ssl_disable:1; */
+    ngx_lua_kong_ssl_ctx_t  *ssl_ctx;
+    ngx_uint_t               proxy_ssl_disable; /* unsigned proxy_ssl_disable:1; */
```
chronolaw, line 13: Should we use `ngx_flag_t`?

chronolaw (approved):

stream/src/ngx_stream_lua_kong_module.h
```diff
@@ -5,10 +5,12 @@
 #include <ngx_config.h>
 #include <ngx_core.h>
 #include <ngx_stream.h>
-
+#include <ngx_stream_lua_api.h>
+#include "../../src/ssl/ngx_lua_kong_ssl.h"
 
 typedef struct {
-    ngx_uint_t              proxy_ssl_disable; /* unsigned proxy_ssl_disable:1; */
+    ngx_lua_kong_ssl_ctx_t  *ssl_ctx;
```
fffonion, line 12: as `ssl_ctx` always have same lifecycle as `ngx_stream_lua_kong_ctx_t`, shall we just use the struct instead of pointer as the member? then you only
need to palloc once and free once (https://github.com/Kong/lua-kong-nginx-module/pull/50/files?diff=split&w=1#diff-9b07aee9a137ad53b90b06f49cb9559b5d484c68eabc9bff56b0732bf47cec5cR106)

PR #51, opened 2022-11-02
chore(*): add check of get_request() and the missing ';'

catbro666:
Reason of adding this check: 
Although it's indeed redundent given our phase check. but it gives us extra protection because this a private API from openresty, it doesn't guaranteeded to be unchanged in the future. And also given the severity when it happens, it will likely cause a segfault, instead of a lua vm crach.

### DIFF
--- a/lualib/resty/kong/tls.lua
+++ b/lualib/resty/kong/tls.lua
@@ -71,8 +71,9 @@ if ngx.config.subsystem == "http" then
         end
 
         local r = get_request()
-        -- no need to check if r is nil as phase check above
-        -- already ensured it
+        if not r then
+            error("no request found")
+        end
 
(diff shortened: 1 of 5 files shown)


lualib/resty/kong/tls.lua
```diff
@@ -71,8 +71,9 @@ if ngx.config.subsystem == "http" then
         end
 
         local r = get_request()
-        -- no need to check if r is nil as phase check above
-        -- already ensured it
+        if not r then
```
chronolaw, line 74: It is a repeated pattern. I think we could define a simple function to wrap it, like:
```
local orign_get_request = ngx.get_request
local function get_request()
...
end
```

catbro666: Good idea!

attenuation (approved):

PR #53, opened 2022-11-09
test(*) enable valgrind

catbro666:


valgrind.suppress
```diff
@@ -207,3 +207,50 @@
    fun:ngx_epoll_init
    fun:ngx_event_process_init
 }
+{
```
catbro666, line 210: These two pseudo-leaks are due to the fact that `proxy_ssl_session_reuse` is `on` by default.

catbro666: Therefore the ssl session will be saved by `SSL_SESSION_up_ref`.
Turning `proxy_ssl_session_reuse` to `off`, then these two "leaks" will disappear.

fffonion (approved):

valgrind.suppress
catbro666: Yeah, I also struggled with whether to add this or not. I think tests should not rely on this kind of non-critical directives. Otherwise, we need to keep an eye on this directive every time we add a new test.
Moreover, if the test does need to set `proxy_ssl_session_reuse` to `on`,  it will appear again.

PR #59, opened 2023-01-26
fix(var): patch set_uri_args

samugi:
set_uri_args modifies the $args variable without accessing our patched metatable. For it to work properly we can invalidate our index table for the value "args" whenever the function `set_uri_args` is called. This way the next access to ngx.var.args will miss the index and fetch the real value of $args.

fix for https://github.com/Kong/kong/issues/10080

[KAG-410](https://konghq.atlassian.net/browse/KAG-410)

[KAG-410]: https://konghq.atlassian.net/browse/KAG-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -13,3 +13,14 @@ not_globals = {
 ignore = {
     "6.", -- ignore whitespace warnings
 }
+
+
+globals = {
+    ngx = {
+        req = {
+            set_uri_args = {
+                read_only = false
+            }
+        }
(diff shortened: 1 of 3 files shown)


bungle (approved):
Looks good to me!

PR #65, opened 2023-08-23
feat(error_log): custom error handler

samugi:
define a custom error log handler that allows appending values of ngx variables to the error log

also deifine a new directive: lua_kong_error_log_append_vars which allows providing the variables to append


src/ngx_http_lua_kong_error_log.c
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    return buf;
+}
+
+
+static ngx_int_t
+set_error_log_handler(ngx_http_request_t *r)
+{
+    r->log_handler = ngx_http_log_error_handler;
```
chobits, line 128: replace original r->log_handler,
note that orignal r->log_handler is initialized in https://github.com/search?q=repo%3Anginx%2Fnginx%20r-%3Elog_handler&type=code

chobits:
> static ngx_array_t *error_log_append_var_indexes;

I think we can use standard implementation in nginx, we can get and save these indexes in module cf->[yourmodule conf]->error_log_append_var_indexes in your directive handling function ngx_http_lua_kong_error_log_append_vars() .

Most nginx modules calling ngx_http_get_indexed_variable() saves the index in its module conf, only one module (src/http//modules/ngx_http_userid_filter_module.c) uses global index value.

And also note that if you want to save the indexes to your module conf(ngx_http_lua_kong_loc_conf_t), we need also add ngx_http_lua_kong_merge_loc_conf, but kong module does not define this function, which is weired.

* src/ngx_http_lua_kong_module.c
```
    ngx_http_lua_kong_create_loc_conf,       /* create location configuration */
    NULL                                     /* merge location configuration */
};
```

If you think this work is too trivial, you can ignore it currently. So i just leaves a comment here, not a "review comment"

src/ngx_http_lua_kong_error_log.c
```diff
@@ -0,0 +1,183 @@
...
+static ngx_array_t *error_log_append_var_indexes;
+
+typedef struct {
+    ngx_str_t   name;
+    ngx_uint_t  index;
+} var_elt_t;
+
+
+static u_char *
+ngx_http_log_error_handler(ngx_http_request_t *r, ngx_http_request_t *sr,
```
chronolaw, line 29: Should have  the prefix `ngx_http_lua_kong_`?

src/ngx_http_lua_kong_error_log.c
```diff
@@ -0,0 +1,183 @@
...
+{
+    error_log_append_var_indexes = ngx_array_create(cf->pool, 4, sizeof(var_elt_t));
+    if (error_log_append_var_indexes == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    ngx_str_t *values = cf->args->elts;
+    ngx_uint_t num_elts = cf->args->nelts - 1;
+
+    // values passed to the directive start at values[1]
```
chronolaw, line 165: Should use C style comment.

chronolaw:
When the feature is finished, we need a readme.

chronolaw:
Could we use `ngx_http_complex_value_t`, which may be easier to use, like:
```
lua_kong_error_log_append    "kong_request_id: $kong_request_id,  nginx_version: $nginx_version";

```

src/ngx_http_lua_kong_error_log.c
chronolaw: We should add a comment like `copy from ngx_http_request.c:ngx_http_log_error_handler()`.

src/ngx_http_lua_kong_error_log.c
chronolaw: Yes, I suggest add some comments to describe the code and how to do in the future.

chronolaw:
I think that perhaps we could not copy the code of `ngx_http_log_error_handler()`,
we can implement our own log logic, the apply a patch to nginx source, injecting the function to the end of `ngx_http_log_error_handler()`.

We already did this thing in lua_kong_module.

src/ngx_http_lua_kong_error_log.c
samugi: > ❗ It replaces the nginx default logic and needs to be reviewed carefully.

correct. Agree.

> if nginx official modifies the logic [...] we need to introduce the official modification to be here

I think we could consider any minor updates to the nginx log handler optional to port here (though it would be probably a good idea for consistency) but not necessarily problematic if left out of sync. Unless of course the code used in our log handler becomes deprecated through some breaking changes in nginx.

src/ngx_http_lua_kong_error_log.c
dndx, line 29: This large duplication of code is concerning. Any other ways?

chronolaw: we can implement our own log logic, the apply a patch to nginx source, injecting the function to the end of ngx_http_log_error_handler().

We already did this thing in lua_kong_module.

samugi: an alternative would be patching the original handler in nginx. Initially I went for a whole separate log handler to avoid patching but it appears based on @chronolaw 's comment that we are already doing something similar so I could go ahead and reduce this duplication by just patching our addition onto the nginx implementation if it feels safer that way.

src/ngx_http_lua_kong_error_log.c
```diff
@@ -0,0 +1,183 @@
...
+        p = ngx_snprintf(buf, len, ", referrer: \"%V\"",
+                         &r->headers_in.referer->value);
+        len -= p - buf;
+        buf = p;
+    }
+
+
+    // loop through error_log_append_var_indexes and append
+    // each variable and their values to the error log
+    var_elt_t *elts  = error_log_append_var_indexes->elts;
```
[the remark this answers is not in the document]
samugi: Fixing this everywhere except this funciton, since we are planning to remove it and patch it elsewhere, so I can keep my modifications in a single block in the meantime

src/ngx_http_lua_kong_error_log.c
samugi: closing as this code no longer exists due to the patching approach (see other comments)

src/ngx_http_lua_kong_error_log.c
samugi: done.

samugi:
@chronolaw @chobits I resolved all the items discussed above, could you take another look?

About tests:

- The build/test workflow appears to have some problem (even in the other open PR), so the tests are currently not running on CI
- I added tests for the newly added feature, which I ran locally and made sure they pass
- I disabled all the tests before pushing, because in order to run them we require Nginx to be patched first, to enable this feature in the Nginx error log handler as discussed above. My plan would be:

1. Fix CI to make sure there are no regressions
2. Merge this if/when green
3. Patch Nginx in Kong master
4. Enable these tests (`009-error-log-append.t`) with another PR afterwards

Let me know what you think, thanks!

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include "ngx_http_lua_kong_common.h"
+
+
+void
+ngx_http_lua_kong_error_handler(ngx_http_request_t *r, u_char **buf, size_t *len)
+{
+    u_char                       *p;
+    ngx_http_lua_kong_loc_conf_t *lcf;
+    ngx_http_complex_value_t     *cpx_ela;
```
chronolaw, line 25: `cpx_ela` is hard to understand, could we use meaningful name instead of acronym？

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,42 @@
...
+    ngx_str_t                     ela;
+
+    lcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_kong_module);
+    cpx_ela = lcf->error_log_append;
+
+    if (cpx_ela != NULL && cpx_ela != NGX_CONF_UNSET_PTR
```
chronolaw, line 31: Could it be a early return？ like:
```
if (cpx_ela == NULL ||  cpx_ela == NGX_CONF_UNSET_PTR) {
  return;
}

if (ngx_http_complex_value(r, cpx_ela, &ela) != NGX_OK) {
  return;
}
```

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,42 @@
...
+        && ngx_http_complex_value(r, cpx_ela, &ela) == NGX_OK
+        && ela.len > 0) {
+
+       p = ngx_snprintf(*buf, len, ", %v", &ela);
```
chronolaw, line 35: Should be `%V`?

src/ngx_http_lua_kong_module.c
```diff
@@ -128,7 +136,22 @@ ngx_http_lua_kong_create_loc_conf(ngx_conf_t* cf)
         return NULL;
     }
 
+    conf->error_log_append = NGX_CONF_UNSET_PTR;
```
chronolaw, line 139: Do we really need it? Is `NULL` not enough? I'm not sure about it.

src/ngx_http_lua_kong_module.c
```diff
@@ -20,6 +20,7 @@
 
 static ngx_int_t ngx_http_lua_kong_init(ngx_conf_t *cf);
 static void* ngx_http_lua_kong_create_loc_conf(ngx_conf_t* cf);
+static char* ngx_http_lua_kong_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child);
```
chronolaw, line 23: I think we need not to enable `merge` feature. Defining in `location` is enough.

src/ngx_http_lua_kong_module.c
```diff
@@ -128,7 +136,22 @@ ngx_http_lua_kong_create_loc_conf(ngx_conf_t* cf)
...
 }
 
 
+static char*
+ngx_http_lua_kong_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
+{
+    ngx_http_lua_kong_loc_conf_t *prev = parent;
+    ngx_http_lua_kong_loc_conf_t *conf = child;
+
+    ngx_conf_merge_str_value(conf->tag, prev->tag, "");
```
chronolaw, line 151: `tag` is location only, should not merge it.

t/009-error-log-append.t
```diff
@@ -0,0 +1,211 @@
+use Test::Nginx::Socket::Lua;
+
+#worker_connections(1014);
+#master_process_enabled(1);
+#log_level('warn');
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 5) -2;
```
chronolaw, line 9:
```suggestion
plan tests => repeat_each() * (blocks() * 5) - 2;
```

t/009-error-log-append.t
```diff
@@ -0,0 +1,211 @@
...
+#no_diff();
+no_long_string();
+#master_on();
+#workers(2);
+check_accum_error_log();
+run_tests();
+
+__DATA__
+
+=== TEST 0: Dummy: remove this test and the SKIP section from all tests below
```
chronolaw, line 20: Should we start the test cases from `1`?

src/ngx_http_lua_kong_log_handler.c
samugi: yes. In an earlier implementation I was passing `ngx_http_variable_value_t` instead so this was a leftover from that. Surprisingly it works both ways.

src/ngx_http_lua_kong_module.c
samugi: it works both ways but I think this is more consistent with other modules

src/ngx_http_lua_kong_module.c
samugi: I think being able to define this directive on the http/server block would be an added value, otherwise we would only be able to provide a correctly formatted error log after the rewrite phase has completed (in location), am I right? And for that to work we'd need the `merge`. At the same time being able to set this in the location block provides added flexibility in case different formats are needed in different locations so I think we should leave all the scopes available. Let me know if you think this makes sense.

t/009-error-log-append.t
samugi: we should indeed, but this first test case is a dummy, which will be removed once we can actually run the rest of the tests (see description). 

It is required otherwise running the file with no tests enabled produces a failure.

src/ngx_http_lua_kong_module.c
samugi: yes! thank you for pointing that out.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,48 @@
...
+
+    if (log_append == NULL || log_append == NGX_CONF_UNSET_PTR) {
+        return;
+    }
+
+    if (ngx_http_complex_value(r, log_append, &log_append_str) != NGX_OK) {
+        return;
+    }
+
+    if (log_append_str.len > 0) {
```
chronolaw, line 39:
```suggestion
    if (log_append_str.len ) {
```
follow Nginx's code style. in `ngx_http_log_error_handler()`

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,48 @@
...
+
+    if (log_append == NULL || log_append == NGX_CONF_UNSET_PTR) {
+        return;
+    }
+
+    if (ngx_http_complex_value(r, log_append, &log_append_str) != NGX_OK) {
+        return;
+    }
+
+    if (log_append_str.len ) {
```
chronolaw, line 39:
```suggestion
    if (log_append_str.len) {
```

README.md
```diff
@@ -140,6 +141,26 @@ you will always get the value where your Lua code runs in but not others.
 
 [Back to TOC](#table-of-contents)
 
+lua\_kong\_error\_log\_append
+-------------------------------------------
+**syntax:** *lua_kong_error_log_append expression;*
```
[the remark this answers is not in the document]
samugi: changed to `value`

chronolaw:
After this PR is merged, I think that we should add it to stream subsystem too.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,48 @@
...
+    if (log_append == NULL || log_append == NGX_CONF_UNSET_PTR) {
+        return;
+    }
+
+    if (ngx_http_complex_value(r, log_append, &log_append_str) != NGX_OK) {
+        return;
+    }
+
+    if (log_append_str.len) {
+
```
dndx, line 40:
```suggestion
```

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,48 @@
...
+    ngx_str_t                     log_append_str;
+
+    lcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_kong_module);
+    log_append = lcf->error_log_append;
+
+    if (log_append == NULL || log_append == NGX_CONF_UNSET_PTR) {
+        return;
+    }
+
+    if (ngx_http_complex_value(r, log_append, &log_append_str) != NGX_OK) {
```
dndx, line 35: I am a bit concerned about this usage. `ngx_http_complex_value` allocates the result string buffer on `r->pool`, and since this is a high frequency operation, this could cause each error log to leak a little bit of memory until request pool is destroyed.

Maybe we should not use complex value in the first place, or consider cache the resulted `ngx_str_t` on the module `ctx` and only clears it if the request ID changes.

src/ngx_http_lua_kong_module.c
dndx: Merge should be implemented so that it works with the Nginx config semantics.

src/ngx_http_lua_kong_module.c
```diff
@@ -128,7 +136,21 @@ ngx_http_lua_kong_create_loc_conf(ngx_conf_t* cf)
...
 
 
+static char*
+ngx_http_lua_kong_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
+{
+    ngx_http_lua_kong_loc_conf_t *prev = parent;
+    ngx_http_lua_kong_loc_conf_t *conf = child;
+
+    ngx_conf_merge_ptr_value(conf->error_log_append, prev->error_log_append, NGX_CONF_UNSET_PTR);
+
```
dndx, line 152: In this case the `tag` variable should be merged too, again, I think it was an overlook before.

dndx (changes requested):

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ngx_http_lua_kong_common.h"
+
+
+void
+ngx_http_lua_kong_error_handler(ngx_http_request_t *r, u_char **buf, size_t *len)
```
dndx, line 21: We might not need a patch for this function. Here is what it could be done:

1. In `POST_READ` phase, store `r->log_hander` into `ctx->original_log_hander`. The type should be a [ngx_http_log_handler_pt](https://lxr.nginx.org/source/xref/nginx/src/http/ngx_http.h?r=9021%3A8d0753760546&fi=ngx_http_log_handler_pt#ngx_http_log_handler_pt).
2. In here, do:
```C
p = ctx->original_log_hander;

/* remaining code that appends request ID */
```

Does this seems to work? This way we might be able to do it with 0 patches to Nginx core which is always better.

dndx (changes requested):

src/ngx_http_lua_kong_module.c
chronolaw: `tag` is `NGX_HTTP_LOC_CONF`, it will not appear in `http` or `server` scope.

src/ngx_http_lua_kong_log_handler.c
chronolaw: The value of complex variable may be different in different nginx phase, cacheing it may not be a good idea.

src/ngx_http_lua_kong_log_handler.c
chronolaw: It means that we will replace the original `r->log_handler` to our own handler.

samugi: I think @dndx means to store the original log handler (in ctx) and then replace it with a function that executes the original + additional code which appends the request ID. If that work that would in fact require no additional patching. I will test that.

src/ngx_http_lua_kong_log_handler.c
samugi: Using the complex value was [a suggestion](https://github.com/Kong/lua-kong-nginx-module/pull/65#issuecomment-1691068556) to replace the original implementation which was taking an array of variable names. It simplifies the code quite a lot, but please let me know if you think there is a safer way to handle this @dndx @chronolaw

src/ngx_http_lua_kong_module.c
dndx: Ah, okay I didn't notice that. Maybe we should add a comment like:

```
/* conf->tag is NGX_HTTP_LOC_CONF only */
```

src/ngx_http_lua_kong_log_handler.c
samugi: here's my proposal, following @dndx's suggestion above (it works!) --> https://github.com/Kong/lua-kong-nginx-module/pull/65/commits/d185ee8abb8032e6cc2b75c4a3b97abddcdc71a1

src/ngx_http_lua_kong_log_handler.c
samugi: in the latest commit I simplified the directive's scope to only set the request ID as suggested, and got rid of the complex value

samugi:
In the meantime, the effort for making master's CI green is tracked [in this PR](https://github.com/Kong/lua-kong-nginx-module/pull/66), however [Valgrind seems to be detecting a leak ](https://github.com/Kong/lua-kong-nginx-module/actions/runs/6013213807/job/16311805698?pr=66#step:9:500) so tests are still failing.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,124 @@
...
+    if (value[1].data[0] != '$') {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                            "invalid variable name \"%V\"", &value[1]);
+        return NGX_CONF_ERROR;
+    }
+    value[1].len--;
+    value[1].data++;
+
+    p = conf;
+    index = (ngx_uint_t *) (p + cmd->offset);
```
chobits, line 115: you can get index from (ngx_http_lua_kong_loc_conf_t *) pointer lcf converted from (void*)conf

```
lcf = conf
...
/* use lcf-> req_id_var_index */
```

t/009-error-log-append.t
```diff
@@ -0,0 +1,171 @@
...
+#no_diff();
+no_long_string();
+#master_on();
+#workers(2);
+check_accum_error_log();
+run_tests();
+
+__DATA__
+
+
```
chobits, line 20: need to add one test case test whether the request_id changed during rewrite or access phase, may be

```

        lua_kong_error_log_request_id $my_var;
        set $my_var "";

         rewrite_by_lua_block {
         ngx.log(ngx.ERR, "...") -- check whether it is empty
         ngx.var.my_var = ..;      -- modify request_id variable
         ngx.log(ngx.ERR, "...")  --  check whether request_id has been changed
         }

         access_by_lua_block {
           ngx.log(ngx.ERR, "...")  --  check whether request_id has been logged during access phase
         }

```

chobits (approved):

src/ngx_http_lua_kong_module.h
```diff
@@ -36,5 +36,8 @@ ngx_flag_t
 ngx_http_lua_kong_get_upstream_ssl_verify(ngx_http_request_t *r,
     ngx_flag_t proxy_ssl_verify);
 
+void
+ngx_http_lua_kong_error_handler(ngx_http_request_t *r,
```
chronolaw, line 40: Do we need it now?

samugi: no, thanks

t/009-error-log-append.t
samugi: added one test case with what should be all the possible changes the request ID can have

.github/workflows/tests.yml
samugi, line 1: we could create a separate PR for this if needed

chobits (approved):

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,120 @@
...
+char *
+ngx_http_lua_kong_error_log_request_id(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    ngx_str_t                    *value;
+    ngx_http_lua_kong_loc_conf_t *lcf = conf;
+
+    value = cf->args->elts;
+    if (value[1].data[0] != '$') {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                            "invalid variable name \"%V\"", &value[1]);
```
chronolaw, line 106: style issue, should aligin with the first parameter.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,120 @@
...
+static u_char *
+ngx_http_lua_kong_error_log_handler(ngx_http_request_t *r, u_char *buf, size_t len)
+{
+    ngx_int_t                    req_id_var_index;
+    ngx_http_variable_value_t    *value;
+    ngx_http_lua_kong_loc_conf_t *lcf;
+
+    lcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_kong_module);
+    req_id_var_index = lcf->req_id_var_index;
+    if (req_id_var_index == NGX_CONF_UNSET) {
```
chronolaw, line 29: `req_id_var_index` seems to be unnecessary, we can use `lcf->req_id_var_index` directly.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,120 @@
...
+ngx_http_lua_kong_replace_error_log_handler(ngx_http_request_t *r)
+{
+    ngx_http_lua_kong_ctx_t      *ctx;
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx == NULL) {
+        return NGX_ERROR;
+    }
+
+    ctx->orig_log_handler = r->log_handler;
```
chronolaw, line 73: Could we add some comments about the calling chain of log?

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,120 @@
...
+    r->log_handler = ngx_http_lua_kong_combined_error_log_handler;
+
+    return NGX_DECLINED;
+}
+
+
+char *
+ngx_http_lua_kong_set_post_read_handler(ngx_conf_t *cf)
```
chronolaw, line 81: How about `ngx_http_lua_kong_error_log_init`, similar as `ngx_lua_kong_ssl_init`

t/009-error-log-append.t
```diff
@@ -0,0 +1,211 @@
+use Test::Nginx::Socket::Lua;
+
+#worker_connections(1014);
+#master_process_enabled(1);
+#log_level('warn');
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 6) - 6;
```
chronolaw, line 9: maybe `repeat_each() * (blocks() * 5)`?

t/009-error-log-append.t
```diff
@@ -0,0 +1,211 @@
...
+no_long_string();
+#master_on();
+#workers(2);
+check_accum_error_log();
+run_tests();
+
+__DATA__
+
+
+=== TEST 1: value is appended correctly to error logs
```
chronolaw, line 21: should we have a "sanity case", which means that if no `lua_kong_error_log_request_id` the error log will not output string "kong_request_id"

src/ngx_http_lua_kong_module.c
```diff
@@ -76,6 +84,7 @@ ngx_module_t ngx_http_lua_kong_module = {
 static ngx_int_t
 ngx_http_lua_kong_init(ngx_conf_t *cf)
 {
+    ngx_http_lua_kong_error_log_init(cf);
```
chronolaw, line 87: We should check the return value of `ngx_http_lua_kong_error_log_init`.

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,129 @@
...
+ */
+static u_char *
+ngx_http_lua_kong_combined_error_log_handler(ngx_http_request_t *r, ngx_http_request_t *sr,
+    u_char *buf, size_t len)
+{
+    u_char                       *p;
+    ngx_http_lua_kong_ctx_t      *ctx;
+
+    ctx = ngx_http_lua_kong_get_module_ctx(r);
+    if (ctx != NULL) {
```
chronolaw, line 55: I have a question: 
if we don't set the directive `lua_kong_error_log_request_id`,
do we pass the original log handler calling?
Does it means we will not log the normal http message?

samugi: no we don't skip the original log handler calling, when the directive is unset we still execute the combined error log handler but the `ngx_http_lua_kong_error_log_handler` [has no effect](https://github.com/Kong/lua-kong-nginx-module/blob/2def18c97bc1521d012f1cff714ddf707f887110/src/ngx_http_lua_kong_log_handler.c#L29) since there is no request id variable index set.

I added [a check in the sanity test case](https://github.com/Kong/lua-kong-nginx-module/blob/ed6732b9af5f8c1ac06dfc81b37e18660a649551/t/009-error-log-append.t#L35) to verify this.

src/ngx_http_lua_kong_module.c
samugi: done, can you check again? thanks

chronolaw (approved):
Looks great, thanks for your effort.

samugi:
now testing this in Kong, there seems to be a problem when executing [this line](https://github.com/Kong/lua-kong-nginx-module/blob/904f3c1f0567fd28ea1c21057feaf7ebf802df3d/src/ngx_http_lua_kong_log_handler.c#L60) in grpc proxy mode, looking into it.

samugi:
successful test run in Kong [using this PR's head](https://github.com/Kong/kong/blob/70900ca780b295ba360117ca443a4a61f21d6f2c/.requirements#L9) as version of the kong nginx module [here](https://github.com/Kong/kong/actions/runs/6084268284/job/16506269639?pr=11308).

src/ngx_http_lua_kong_log_handler.c
```diff
@@ -0,0 +1,139 @@
...
+        return buf;
+    }
+
+    buf = ngx_snprintf(buf, len, ", kong_request_id: %v", value);
+    return buf;
+}
+
+
+/*
+ * This function is a replacement for the original nginx error
```
dndx, line 43:
```suggestion
 * This function replaces the original HTTP error
```

dndx (commented):
Mostly just nitpicks, but overall this implementation looks safe to my eyes.

dndx (approved):
LGTM, very good work indeed

PR #70, opened 2023-10-10
feat(src): add new var $kong_request_id

chronolaw:
See: https://konghq.atlassian.net/browse/KAG-2734

After this PR is merged, please bump the module's version.

Related PR: https://github.com/Kong/kong/pull/11725


hanshuebner:
Given that the request ID is created for every request, it may make sense to replace the use of `sprintf` by a specialized hexadecimal serializer.  In my tests, this can reduce the time to generate the request ID string by 80%.  In the grand scheme of things, this may not be too much, but it would be an improvement that is cheap to make:
```
#include <stdlib.h>
#include <stdio.h>

void
hex_string(uint64_t num, char *s)
{
  static const char digits[513] =
    "000102030405060708090a0b0c0d0e0f"
    "101112131415161718191a1b1c1d1e1f"
    "202122232425262728292a2b2c2d2e2f"
    "303132333435363738393a3b3c3d3e3f"
    "404142434445464748494a4b4c4d4e4f"
    "505152535455565758595a5b5c5d5e5f"
    "606162636465666768696a6b6c6d6e6f"
    "707172737475767778797a7b7c7d7e7f"
    "808182838485868788898a8b8c8d8e8f"
    "909192939495969798999a9b9c9d9e9f"
    "a0a1a2a3a4a5a6a7a8a9aaabacadaeaf"
    "b0b1b2b3b4b5b6b7b8b9babbbcbdbebf"
    "c0c1c2c3c4c5c6c7c8c9cacbcccdcecf"
    "d0d1d2d3d4d5d6d7d8d9dadbdcdddedf"
    "e0e1e2e3e4e5e6e7e8e9eaebecedeeef"
    "f0f1f2f3f4f5f6f7f8f9fafbfcfdfeff";
  uint32_t x = (uint32_t)num;
  int i = 3;
  while (i >= 0) {
    int pos = (x & 0xFF) * 2;
    char ch = digits[pos];
    s[i * 2] = ch;

    ch = digits[pos + 1];
    s[i * 2 + 1] = ch;

    x >>= 8;
    i -= 1;
  }
}


int
main()
{
  char buf[37];
  for (int i = 0; i < 10000000; i++) {
#if FAST
    hex_string(rand(), buf);
    buf[8] = 'D';
    hex_string(rand(), buf+9);
    buf[17] = 'D';
    hex_string(rand(), buf+18);
    buf[26] = 'D';
    hex_string(rand(), buf+27);
    buf[35] = 'D';
    buf[36] = 0;
#else
    sprintf(buf, "%08xD%08xD%08xD%08xD", rand(), rand(), rand(), rand());
#endif
  }

  return 0;
}
```

hanshuebner (changes requested):

src/ngx_http_lua_kong_request_id.c
```diff
@@ -0,0 +1,83 @@
...
+    if (id == NULL) {
+        return NGX_ERROR;
+    }
+
+    v->valid = 1;
+    v->no_cacheable = 0;
+    v->not_found = 0;
+
+    v->len = KONG_REQUEST_ID_MAX_LEN;
+    v->data = id;
```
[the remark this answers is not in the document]
chronolaw: If we don't set the flag `NGX_HTTP_VAR_CHANGEABLE`, it will be unchangeable by default.

src/ngx_http_lua_kong_request_id.c
```diff
@@ -0,0 +1,83 @@
...
+
+    ngx_sprintf(id, KONG_REQUEST_ID_FORMAT,
+                (uint32_t) ngx_random(), (uint32_t) ngx_random(),
+                (uint32_t) ngx_random(), (uint32_t) ngx_random());
```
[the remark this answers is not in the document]
chronolaw: I think that we can use `ngx_hex_dump()` in nginx code base.

hanshuebner (dismissed):

hanshuebner (changes requested):
Please provide some documentation - Either in the pull request description or the README.  The code looks good to me.

README.md
```diff
@@ -160,6 +162,16 @@ An error log line may look similar to the following:
 
 [Back to TOC](#table-of-contents)
 
+Variables
+=========
+
+$kong\_reques\_id
```
flrgh, line 168:
```suggestion
$kong\_request\_id
```

zhongweiy:
The code looks good to me.

README.md
chronolaw: thanks, done.

chronolaw:
> Have we seeded the `ngx_random` in other place? `ngx_random` calls libc `random` if I remember correctly and if we do not seed it, it will be seeded with `1` according to https://man7.org/linux/man-pages/man3/random.3.html.
> 
> If it has been seeded, the code looks good to me.

Here it is:
```
os/unix/ngx_process_cycle.c:    srandom(((unsigned) ngx_pid << 16) ^ tp->sec ^ tp->msec);
os/unix/ngx_posix_init.c:    srandom(((unsigned) ngx_pid << 16) ^ tp->sec ^ tp->msec);
```

ADD-SP (commented):
I think almost all changes are good except the length of `$kong_request_id`.

Currently, we use 8 bytes as `$kong_request_id`, the collision chance is `sqrt(pow(256, 8)) = 4,294,967,296 (uint32_t)`.

My concern is about the cloud scenario, if there is a large volunm of traffic, the `$kong_request_id` might replicate many times in a short of period.

t/010-request-id.t
```diff
@@ -0,0 +1,42 @@
...
+run_tests();
+
+__DATA__
+
+=== TEST 1: $kong_request_id works
+--- config
+    location /t {
+        content_by_lua_block {
+            local rid = ngx.var.kong_request_id
+            assert(ngx.re.match(rid, "[0-9a-f]{16}"))
```
zhongweiy, line 26: Should the rid a 32 haxadecimal characters string instead of 16?

chronolaw: Now we are using 2 uint32_t random number, two numbers cause 8 bytes, so the hex result is 16 bytes.

chronolaw:
> I think almost all changes are good except the length of `$kong_request_id`.
> 
> Currently, we use 8 bytes as `$kong_request_id`, the collision chance is `sqrt(pow(256, 8)) = 4,294,967,296 (uint32_t)`.
> 
> My concern is about the cloud scenario, if there is a large volunm of traffic, the `$kong_request_id` might replicate many times in a short of period.

Yes, it needs more discussion, Perhaps 24 bytes is a choice.

ADD-SP (commented):
I think we can just do what the Nginx built-in `$request_id` did, using 16 bytes random data as `$kong_request_id`.

chronolaw:
> I think we can just do what the Nginx built-in `$request_id` did, using 16 bytes random data as `$kong_request_id`.

Of course, but if 8 bytes random data is enough we could get a little speed up.

src/ngx_http_lua_kong_vars.c
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include "ngx_http_lua_kong_common.h"
+
+
+#define RANDOM_COUNT         2
+#define UINT32_HEX_LEN       sizeof(uint32_t) * 2
```
dndx, line 22: Prefix these with `NGX_HTTP_LUA_KONG` to avoid contaminating the global namespace.

config
```diff
@@ -3,11 +3,12 @@ ngx_module_name=ngx_http_lua_kong_module
 ngx_module_srcs=" \
                 $ngx_addon_dir/src/ngx_http_lua_kong_grpc.c \
                 $ngx_addon_dir/src/ngx_http_lua_kong_ssl.c \
-                $ngx_addon_dir/src/ngx_http_lua_kong_var.c \
+                $ngx_addon_dir/src/ngx_http_lua_kong_var_idx.c \
```
dndx, line 6:
```suggestion
                $ngx_addon_dir/src/ngx_http_lua_kong_var_index.c \
```

Do not abbreviate filenames and stay consistent with the name convention.

dndx (changes requested):

src/ngx_http_lua_kong_vars.c
```diff
@@ -0,0 +1,83 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include "ngx_http_lua_kong_common.h"
+
+
+#define NGX_HTTP_LUA_KONG_RANDOM_COUNT         2
```
dndx, line 21: I suggest we use 16 bytes, same as `$request_id`. We can not deviate from the original design of the feature.

dndx (changes requested):

src/ngx_http_lua_kong_vars.c
chronolaw: done.

t/010-request-id.t
chronolaw: Now it is 16 bytes random bytes ,32 haxadecimal characters string.

chobits (approved):

README.md
```diff
@@ -160,6 +162,16 @@ An error log line may look similar to the following:
 
 [Back to TOC](#table-of-contents)
 
+Variables
+=========
+
+$kong\_request\_id
+------------------
+Unique request identifier generated from 16 random bytes, in hexadecimal.
+This variable has already been indexed.
```
hanshuebner, line 171:
```suggestion
This variable is indexed.
```

t/010-request-id.t
```diff
@@ -0,0 +1,42 @@
...
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 5);
+
+my $pwd = cwd();
+
+$ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
+
+no_long_string();
+#no_diff();
```
hanshuebner, line 15: Can this go?

README.md
```diff
@@ -160,6 +162,16 @@ An error log line may look similar to the following:
 
 [Back to TOC](#table-of-contents)
 
+Variables
+=========
+
+$kong\_request\_id
+------------------
+Unique request identifier generated from 16 random bytes, in hexadecimal.
```
hanshuebner, line 170:
```suggestion
Unique request identifier generated from 16 pseudo-random bytes, in hexadecimal.
```

hanshuebner (changes requested):

hanshuebner (approved):

samugi (approved):

ADD-SP (approved):

zhongweiy (approved):

dndx (approved):

PR #72, opened 2023-10-12
feat(var): add three variables that are indexed by default

chronolaw:
See: https://github.com/Kong/kong/pull/11722#discussion_r1354048422


ADD-SP (approved):

ADD-SP (changes requested):
Please add tests into the `t/006-default_indexed-var.t`.

ADD-SP (approved):

PR #76, opened 2024-01-11
chore(ci): test for openresty 1.25.3.1

chronolaw:
Related PR : https://github.com/Kong/kong/pull/12327


chronolaw:
mockbin.com is deprecated, we should updated the test cases.

PR #97, opened 2024-12-11
fix(var): patch set_header

tzssangglass:
patch the `req.set_header` function to invalidate the relevant `variable_index` entry for the modified header. Specifically, after normalizing the header name, the corresponding `variable_index` entry is set to `nil`. This ensures that subsequent accesses to `ngx.var.http_*` variables will bypass the cached index and fetch the updated header value.

same fix way as: https://github.com/Kong/lua-kong-nginx-module/pull/59

Fix: [KAG-5963](https://konghq.atlassian.net/browse/KAG-5963)
Fix: [FTI-6406](https://konghq.atlassian.net/browse/FTI-6406)

[KAG-5963]: https://konghq.atlassian.net/browse/KAG-5963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FTI-6406]: https://konghq.atlassian.net/browse/FTI-6406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lualib/resty/kong/var.lua
+++ b/lualib/resty/kong/var.lua
@@ -24,6 +24,8 @@ local NGX_DECLINED = ngx.DECLINED
 
 local variable_index = {}
 local metatable_patched
+local str_replace_char
+local replace_dashes_lower
 
 
 --Add back if stream module is implemented to aid readability
@@ -50,6 +52,11 @@ if subsystem == "http" then
     --ngx_lua_kong_ffi_var_get_by_index = C.ngx_http_lua_kong_ffi_var_get_by_index
     --ngx_lua_kong_ffi_var_set_by_index = C.ngx_http_lua_kong_ffi_var_set_by_index
     --ngx_lua_kong_ffi_var_load_indexes = C.ngx_http_lua_kong_ffi_var_load_indexes
+    
+    str_replace_char = require("resty.core.utils").str_replace_char
+    replace_dashes_lower = function(str)
+        return str_replace_char(str:lower(), "-", "_")
+    end
 end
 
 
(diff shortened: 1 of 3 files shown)


lualib/resty/kong/var.lua
```diff
@@ -159,6 +166,16 @@ local function patch_functions()
     variable_index.args = nil
     return orig_set_uri_args(...)
   end
+  
+  local orig_set_header = req.set_header
+
+  req.set_header = function(...)
+    local normalized_header = replace_dashes_lower(...)
```
tzssangglass, line 173: perhaps I need to unpack the data structure(…) and retrieve the first element; this should be clear.

samugi: right, doesn't the current implementation mean we're passing both `header_name` and `header_value` from `ngx.req.set_header(header_name, header_value)`, to `replace_dashes_lower`?

We can probably patch this with a function that takes exactly 2 arguments, then use the first (header_name) to invalidate the index.

tzssangglass: I changed to
 
```lua
  req.set_header = function(name, value)
    local normalized_header = replace_dashes_lower(name)
```

 I did not choose the

```lua
  req.set_header = function(...)
    local args = table.unpack(...)
    local normalized_header = replace_dashes_lower(args[1])
```

I feel that this will degrade performance.

samugi (approved):

ms2008 (approved):

outsinre (approved):

catbro666 (approved):

PR #110, opened 2025-05-15
feat(tls): add SANs support for both DNS-names and URIs

lucab:
This adds two new Lua functions (and the corresponding C logic) to configure custom SANs validation in nginx:
 * `resty.kong.tls.set_upstream_ssl_sans_dnsnames()`
 * `resty.kong.tls.set_upstream_ssl_sans_uris()`

Part of KAG-6247.

---

Notes:
 * This requires a corresponding nginx patch implementing the additional TLS-name validation logic.
 * Testing for this change (and the above nginx patch) is done at the integration level in Kong, see linked PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Table of Contents
     * [resty.kong.tls.disable\_session\_reuse](#restykongtlsdisable_session_reuse)
     * [resty.kong.tls.get\_full\_client\_certificate\_chain](#restykongtlsget_full_client_certificate_chain)
     * [resty.kong.tls.set\_upstream\_cert\_and\_key](#restykongtlsset_upstream_cert_and_key)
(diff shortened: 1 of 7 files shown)


samugi (approved):
lgtm. Is this testable? Should we have something in `t/`?

lucab:
> Is this testable? Should we have something in t/?

I think this is not testable in isolation (at least not in a meaningful way),  as it works in tandem with the related nginx patch wrt. TLS-name validation logic.
However, the composite result of those patches is covered by a new integration test in Kong.

lucab:
@bungle @samugi Thanks for the reviews! Two additional notes on this:
 * I just realized the new function were missing from the README. I added a commit on top with doc entries only.
 * Even if green, I can't self-merge in this repository. Can you please merge this for me?

samugi (approved):

PR #112, opened 2025-09-18
fix(vars): use strong random value to construct request id

zhikai-kong:
We found dup kong_request_id in customer environment. The reason for the dup request id is caused by the random seed in nginx:

srandom(((unsigned) ngx_pid << 16) ^ tp->sec ^ tp->msec);

The seed is not safe if we run kong in different pods. Because the ngx_pid might be the same. If so, the entrophy from the seed might not be enough.

Changed to use RAND_bytes from openssl for generating the request id.

### DIFF
--- a/src/ngx_http_lua_kong_module.c
+++ b/src/ngx_http_lua_kong_module.c
@@ -19,6 +19,7 @@
 #include "ngx_http_upstream.h"
 
 static ngx_int_t ngx_http_lua_kong_init(ngx_conf_t *cf);
+static ngx_int_t ngx_http_lua_kong_init_worker(ngx_cycle_t *cycle);
 static void* ngx_http_lua_kong_create_loc_conf(ngx_conf_t* cf);
 static char* ngx_http_lua_kong_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child);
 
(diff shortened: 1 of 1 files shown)


chronolaw:
It may have performance impact. We must be careful about it.

chronolaw:
See https://konghq.atlassian.net/browse/KAG-2692

src/ngx_http_lua_kong_vars.c
```diff
@@ -45,7 +45,11 @@ ngx_http_lua_kong_variable_request_id(ngx_http_request_t *r,
     v->data = id;
 
     for (i = 0; i < NGX_HTTP_LUA_KONG_RANDOM_COUNT; i++) {
-        rnd = (uint32_t) ngx_random();
+        if (RAND_bytes((u_char *)&rnd, sizeof(rnd)) <= 0) {
```
chronolaw, line 48: We should not use `RAND_bytes`, it is too slow.

chronolaw (changes requested):

src/ngx_http_lua_kong_module.c
```diff
@@ -19,6 +19,7 @@
 #include "ngx_http_upstream.h"
 
 static ngx_int_t ngx_http_lua_kong_init(ngx_conf_t *cf);
+static ngx_int_t ngx_lua_kong_reset_random_seed(ngx_conf_t *cf);
```
chronolaw, line 22: could be `ngx_http_lua_kong_random_seed`?

chronolaw:
Should we create a patch for `ngx_process_cycle.c`? It may be more efficient.

zhikai-kong:
We have decided to patch nginx: https://github.com/Kong/kong-ee/pull/14224
Close this one.

PR #117, opened 2026-03-25
feat(vars): add worker level connection counts

Icarus9913:
### Summary

Add worker level connection counts as variable.

### DIFF
--- a/src/ngx_http_lua_kong_var_index.c
+++ b/src/ngx_http_lua_kong_var_index.c
@@ -102,6 +102,8 @@ static ngx_str_t default_vars[] = {
 
     /* lua-kong-module vars */
     ngx_string("kong_request_id"),
+    ngx_string("kong_worker_connections_total"),
+    ngx_string("kong_worker_connections_free"),
 
     ngx_null_string
 };
--- a/src/ngx_http_lua_kong_var_index.c
+++ b/src/ngx_http_lua_kong_var_index.c
@@ -102,7 +102,6 @@ static ngx_str_t default_vars[] = {
 
     /* lua-kong-module vars */
     ngx_string("kong_request_id"),
-    ngx_string("kong_worker_connections_total"),
     ngx_string("kong_worker_connections_free"),
 
     ngx_null_string
--- a/src/ngx_http_lua_kong_vars.c
+++ b/src/ngx_http_lua_kong_vars.c
@@ -211,6 +211,48 @@ not_found:
 #endif /* NGX_SSL */
 
 
+static ngx_int_t
+ngx_http_lua_kong_variable_worker_connections_total(ngx_http_request_t *r,
+    ngx_http_variable_value_t *v, uintptr_t data)
+{
+    u_char  *p;
+
+    p = ngx_pnalloc(r->pool, NGX_INT_T_LEN);
+    if (p == NULL) {
+        return NGX_ERROR;
+    }
+
+    v->len = ngx_sprintf(p, "%ui", ngx_cycle->connection_n) - p;
+    v->valid = 1;
+    v->no_cacheable = 0;
+    v->not_found = 0;
+    v->data = p;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_lua_kong_variable_worker_connections_free(ngx_http_request_t *r,
+    ngx_http_variable_value_t *v, uintptr_t data)
+{
+    u_char  *p;
+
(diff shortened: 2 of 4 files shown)


src/ngx_http_lua_kong_vars.c
```diff
@@ -211,6 +211,48 @@ ngx_http_lua_kong_get_upstream_tls_protocol(ngx_http_request_t *r, ngx_http_vari
...
+    u_char  *p;
+
+    p = ngx_pnalloc(r->pool, NGX_INT_T_LEN);
+    if (p == NULL) {
+        return NGX_ERROR;
+    }
+
+    v->len = ngx_sprintf(p, "%ui", ngx_cycle->free_connection_n) - p;
+    v->valid = 1;
+    v->no_cacheable = 1;
```
iykon, line 227: The [indexed getter](https://github.com/Kong/lua-kong-nginx-module/blob/master/src/ngx_http_lua_kong_var_index.c#L208) doesn't respect [no_cacheable](https://github.com/Kong/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_variables.c#L622-L624)
It should probably use [ngx_http_get_flushed_variable](https://github.com/Kong/nginx/blob/2485681308bd8d3108da31546cb91bb97813a3fb/src/http/ngx_http_variables.c#L658-L659) which respects no_cacheble.

iykon: maybe add a test case:
get free connection twice and change the connection in between

src/ngx_http_lua_kong_vars.c
```diff
@@ -211,6 +211,48 @@ ngx_http_lua_kong_get_upstream_tls_protocol(ngx_http_request_t *r, ngx_http_vari
 #endif /* NGX_SSL */
 
 
+static ngx_int_t
```
iykon: do you need to add these to https://github.com/Kong/lua-kong-nginx-module/blob/master/src/ngx_http_lua_kong_var_index.c#L22?

src/ngx_http_lua_kong_vars.c
chobits: I found that @Icarus9913 just rebased the commits and it has included the test case.

chobits: it seems that we should call 

```
ngx_http_variable_value_t *
ngx_http_get_flushed_variable(ngx_http_request_t *r, ngx_uint_t index)
{
    ngx_http_variable_value_t  *v;

    v = &r->variables[index];

    if (v->valid || v->not_found) {
        if (!v->no_cacheable) {
            return v;
        }

        v->valid = 0;
        v->not_found = 0;
    }

    return ngx_http_get_indexed_variable(r, index);
}
```

in our [ngx_http_lua_kong_vars.c](https://github.com/Kong/lua-kong-nginx-module/blob/master/src/ngx_http_lua_kong_var_index.c#L208) to support cachable var or non-cachable one.

but currently i think it 's ok, as every variables are used as non-cachable.

Icarus9913: - `ngx_http_get_indexed_variable`: returns cached result immediately if `r->variables[index].valid == 1` — it never checks no_cacheable
- `ngx_http_get_flushed_variable` checks `valid && no_cacheable → if both true, it flushes `(valid=0)` and re-calls the getter

chobits (approved):

src/ngx_http_lua_kong_var_index.c
```diff
@@ -102,6 +102,8 @@ static ngx_str_t default_vars[] = {
 
     /* lua-kong-module vars */
     ngx_string("kong_request_id"),
+    ngx_string("kong_worker_connections_total"),
```
bungle: As discussed on Slack, Kong already calculates value for:
https://nginx.org/en/docs/ngx_core_module.html#worker_connections

So Kong already knows it and the value can probably even be found in `kong.configuration`. If that value is same as here, we don't need a new API.

bungle (changes requested):
We probably don't need `total` api.

Icarus9913:
Should we rename the variable name from `kong_worker_connections_free` to `kong_worker_free_connection_n` since we extract that value from the C codes?

```text
 ngx_cycle->free_connection_n
```

@bungle @iykon

iykon (approved):

bungle (approved):

PR #122, opened 2026-08-17
feat(request): add resty.kong.request.had_body() to tell whether a request body was sent

tzssangglass:
Works without relying on the Content-Length header, which HTTP/2 requests lack because the framing layer alone delimits the body length. A zero-length body (e.g., an HTTP/2 zero-length DATA frame with END_STREAM) is reported as no body.

Detection order: the received-bytes counter maintained by the HTTP/2 body filter, the HTTP/2 preread buffer, stream->in_closed (client done sending with zero bytes), and finally nginx's own
content_length_n/chunked test. The Lua wrapper rejects HTTP/3 for now (it has no pre-read signal to answer before the body is read).

Tests: t/013 (HTTP/1.x) and t/014 (HTTP/2, using Test::Nginx's built-in "--- http2")


src/ngx_http_lua_kong_body.c
```diff
@@ -0,0 +1,86 @@
...
+        && rb->last_saved)
+    {
+        return 0;
+    }
+#endif
+*/
+
+    if (r->headers_in.content_length_n > 0 || r->headers_in.chunked) {
+        return 1;
+    }
```
STARRY-S, line 76: I think there is one HTTP/2 edge case that this PR does not cover.

A client can send HEADERS without END_STREAM, then close the request with an empty DATA frame that carries END_STREAM. The completed request body is empty. However, while the stream is still pending, Nginx sets `r->headers_in.chunked = 1`, and `had_body()` returns `true`. After `ngx.req.read_body()` receives the empty DATA frame, `had_body()` returns `false`.

The current tests cover empty requests where **curl sets END_STREAM on the HEADERS frame**. They **do not cover this delayed empty-stream sequence**.

I opened draft PR [#123](https://github.com/Kong/lua-kong-nginx-module/pull/123) to demonstrate the case. Its tests intentionally expect `false:false:`, while the current implementation returns `true:false:`.

tzssangglass: fixed by: https://github.com/Kong/lua-kong-nginx-module/pull/122/changes/04b3025bbedc5fcc69bd2ce134af8bb6b6176fb2

README.md
```diff
@@ -37,6 +37,8 @@ Table of Contents
     * [resty.kong.log.set\_log\_level](#restykonglogset_log_level)
     * [resty.kong.log.get\_log\_level](#restykonglogget_log_level)
     * [resty.kong.upstream.set\_next\_upstream](#restykongupstreamset_next_upstream)
+    * [resty.kong.request.had\_body](#restykongrequesthad_body)
+    * [resty.kong.request.had\_body](#restykongrequesthad_body)
```
outsinre: Two identical sections?

src/ngx_http_lua_kong_body.c
```diff
@@ -0,0 +1,96 @@
...
+ * it never resolves for HTTP/1.x chunked requests, because nginx keeps no
+ * received-bytes count for HTTP/1.x.
+ *
+ * Order: rb->received > 0 (exact counter, maintained by h2 only) /
+ * h2 preread buffer non-empty / client done sending with zero bytes
+ * (h2: stream->in_closed) / else content_length_n > 0 is proof of a body,
+ * chunked is pending.  HTTP/3 is rejected by the Lua wrapper.
+ */
+ngx_int_t
+ngx_http_lua_kong_req_had_body(ngx_http_request_t *r)
```
outsinre: Not a blocker, but `has_body` may be better.

lualib/resty/kong/request.lua
```diff
@@ -0,0 +1,57 @@
+-- Copyright 2019-2025 Kong Inc.
```
STARRY-S, line 1: Not a blocker: 
Should we update to `2019-2026`?

README.md
tzssangglass: remove it

src/ngx_http_lua_kong_body.c
tzssangglass: change it.

src/ngx_http_lua_kong_body.c
```diff
@@ -0,0 +1,96 @@
...
+#endif
+*/
+
+    if (r->headers_in.content_length_n > 0) {
+        return 1;
+    }
+
+    if (r->headers_in.chunked) {
+        return -1;   /* body framing declared, payload unknown yet */
+    }
```
STARRY-S, line 80: Probably need more logic when `chunked = 1` and the body was read:

We may resolve an empty HTTP/1.x chunked body **after `ngx.req.read_body()` completes**.

Returning `pending` *before the body is read* is correct. However, after a complete buffered read, Nginx sets [`rb->rest` to `0`](https://github.com/nginx/nginx/blob/release-1.29.2/src/http/ngx_http_request_body.c#L1218-L1233) and [`rb->last_saved` to `1`](https://github.com/nginx/nginx/blob/release-1.29.2/src/http/ngx_http_request_body.c#L1309-L1318). Nginx itself [uses both fields to determine that the body read is complete](https://github.com/nginx/nginx/blob/release-1.29.2/src/http/ngx_http_request_body.c#L431-L433). For an empty chunked body, `content_length_n` also remains `0`. At that point, the body is known to be empty, but the current code still returns `-1` because `headers_in.chunked` remains set.

We could check the completed request-body state before the `chunked` fallback. A completed body with a final length of zero should return `false`. The existing empty chunked test after `read_body()` should then expect `false` instead of `pending`.

Example:

```C
if (rb != NULL && rb->rest == 0 && rb->last_saved) {
    return r->headers_in.content_length_n > 0 ? 1 : 0;
}

if (r->headers_in.content_length_n > 0) {
    return 1;
}

if (r->headers_in.chunked) {
    return -1;
}
```

README.md
```diff
@@ -651,6 +652,60 @@ previous ones.
...
+
+Two further notes:
+
+* The answer always describes the downstream client request. Inside a subrequest
+  (`ngx.location.capture`) it still reports the main request, not the body given
+  to the subrequest.
+* Rewriting the body from Lua with
+  [ngx.req.set_body_data](https://github.com/openresty/lua-nginx-module#ngxreqset_body_data)
+  changes the answer, because it rewrites the request's content length the same
+  way it changes `ngx.var.content_length`.
```
STARRY-S: The note about `ngx.req.set_body_data()` does not match the HTTP/2 implementation. `set_body_data()` updates the request buffers and `content_length_n`, but it does not update `rb->received` or `stream->in_closed`. These fields take precedence in `has_body()`, so the result can still describe the original HTTP/2 body.
I suggest that this API only describes the body sent by the client. We should remove the current rewrite claim and document that callers must invoke `has_body()` before any Lua request-body mutation. Supporting the effective body after a rewrite should be a separate API contract.

cv1002 (changes requested):
### 🧹 Drop the commented-out HTTP/3 block

`src/ngx_http_lua_kong_body.c:65-76` — dead `#if (NGX_HTTP_V3)` code inside a comment. Keep the one-line rationale, drop the code.

src/ngx_http_lua_kong_body.c
```diff
@@ -0,0 +1,96 @@
...
+ *
+#if (NGX_HTTP_V3)
+    if (r->http_version == NGX_HTTP_VERSION_30
+        && rb != NULL
+        && rb->last_saved)
+    {
+        return 0;
+    }
+#endif
+*/
```
cv1002: 🧹 Drop this commented-out HTTP/3 block — dead code, kept alive only as a comment. Keep the one-line rationale, remove the code.

```suggestion
/* HTTP/3 has no pre-read signal, so an exact answer needs the body read
 * to its end. Not implemented yet -- the Lua wrapper rejects HTTP/3
 * before this function is reached. */
```

cv1002 (changes requested):
See inline comment.

lualib/resty/kong/request.lua
windmgc: Let's do it in a separate PR

cv1002 (approved):

src/ngx_http_lua_kong_body.c
tzssangglass: remove it

README.md
tzssangglass: Change the desc and add test cases to cover it; Now this API only re-flattens the real origin request body, don't be influenced by the `ngx.req.set_body_data`

src/ngx_http_lua_kong_body.c
tzssangglass: resolve in: https://github.com/Kong/lua-kong-nginx-module/pull/122/changes/ae5e34e1592517983db8fa73806cbd2479ca57a5

STARRY-S (approved):

lualib/resty/kong/request.lua
```diff
@@ -0,0 +1,57 @@
...
+
+function _M.has_body()
+    local r = get_request()
+
+    if not r then
+        error("no request found")
+    end
+
+    if http_version() == 3.0 then
+        error("has_body() does not support HTTP/3 yet")
```
outsinre, line 44: Just wonder if we should `error` or just return a `nil, "..."`.

If we error, then every time we call this method, we should detect the http version.

windmgc (approved):

